PR #1, opened 2015-03-09
glog: release 0.3.4

ukai:
migrate to https://github.com/google/glog
use autoconf 2.69 / automake 1.14.1


pphaneuf:
Checking in the automatically generated files?

ukai:
yeah, it used to check in files generated by autoconf etc.
in old days, it was recommended way? http://www.gnu.org/software/automake/manual/html_node/CVS.html

anyway, I don't object to auto generated file removal.

pphaneuf:
If you're used to doing that, I'd say no rush. I could send you a pull request later on to remove them, and possibly add the traditional ```autogen.sh``` script used to generate them.

ukai:
ok

PR #5, opened 2015-03-23
Adds color support for tmux

darkwind-mt:
Even with 256 color support enabled in tmux, glog previously would only render plain text.  This pull request adds glog color support for 256 color tmux terminals.


darkwind-mt:
I just completed a CLA for my individual 'darkwind-mt' account.  Thanks!

darkwind-mt:
updated my alternative email address

PR #19, opened 2015-04-27
Added fixed log name support

ezegomez:
See https://code.google.com/p/google-glog/issues/detail?id=209 - this is an updated version of that patch.

It adds a flag that allows to switch behavior from base_filename + filename_extension + time_pid_string to base_filename + filename_extension, while still defaulting to the current behavior to avoid breakage in existing code. This change would allow easier log rotation schemes and better control on what's written on disk.

### DIFF
--- a/src/glog/logging.h.in
+++ b/src/glog/logging.h.in
@@ -330,6 +330,9 @@ typedef unsigned __int64 uint64;
   using fLS::FLAGS_##name
 #endif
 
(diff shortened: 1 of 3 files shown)


ezegomez:
I signed it

harlowja:
Can we get this merged?

sergiud:
I'm reverting this PR due to CI failures. Please revise and resubmit.

PR #20, opened 2015-04-30
Update vlog_is_on.cc

romange:
SetVLOGLevel locks vmodule_lock however RAW_VLOG at the end also implicitly locks it via InitVLOG3__.

This causes deadlock/segfault.

[Review on Reviewable](https://reviewable.io/reviews/google/glog/20)


telkomsel13:
#0

Issue #23, opened 2015-05-05
Logfile with more restrictive permissions control


theopolis:
It would be great if we could open files with more restrictive permissions: https://github.com/google/glog/blob/master/src/logging.cc#L900. Opening with `o600` would be safer for daemons logging potentially sensitive WARN/ERROR details.

Something like:

``` cpp
FLAGS_logfile_mode = 0600;
```

Would be :+1:

Issue #22, opened 2015-05-05
Logfile name (YYYYMMDD-HHMMss, etc) settings and control


theopolis:
Enhancement+feature request+question mostly.

It would be nice to accumulate logs into a `appname.INFO.YYYYMMDD` instead of the very specific filenames generated from: https://github.com/google/glog/blob/master/src/logging.cc#L988

There might be an existing control to limit/changes these filename sinks already?

PR #25, opened 2015-05-11
[#23 #22] Allow a logfile mode and logfile date-naming

theopolis:
Two features we've modded into https://github.com/facebook/osquery.git:

1. Write logs that may have sensitive content in warnings/errors as o600 instead of o644.
2. Write/append to a logfilename using the date as a suffix. If folks are pulling these files into an elk-stack or splunk, having a by-date file name from the project is a nicety! 

It would be great to have 1. in maintain so we can build a safe glog dependency in FreeBSD, where we construct our build hosts with ports. 2. is more of a like-to-have-- so I'm interested in other's thoughts.

### DIFF
--- a/src/glog/logging.h.in
+++ b/src/glog/logging.h.in
@@ -360,6 +360,12 @@ DECLARE_int32(minloglevel);
 // default logging directory.
 DECLARE_string(log_dir);
 
+// Set the log file mode.
+DECLARE_int32(logfile_mode);
+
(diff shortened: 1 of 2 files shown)


theopolis:
Yo @googlebot, I signed it!

pphaneuf:
I'd recommend splitting off the date-naming change to its own pull request, and making this pull requests only about the permission?

PR #26, opened 2015-05-13
[#23] Add logfile_mode to control logfile permissions

theopolis:
I'd like to set o600 for daemons writing potentially sensitive WARN/ERROR messages.

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -161,6 +161,8 @@ static const char* DefaultLogDir() {
   return "";
 }
 
+GLOG_DEFINE_int32(logfile_mode, 0664, "Log file mode/permissions.");
+
 GLOG_DEFINE_string(log_dir, DefaultLogDir(),
                    "If specified, logfiles are written into this directory instead "
                    "of the default logging directory.");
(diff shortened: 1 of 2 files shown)


src/logging.cc
```diff
@@ -161,6 +161,8 @@ static const char* DefaultLogDir() {
   return "";
 }
 
+GLOG_DEFINE_int32(logfile_mode, 0644, "Log file mode/permissions");
```
pphaneuf: The default permission is different than before.

I think it was wrong before, though? The "Unix tradition" is to use 0666, and let the `umask` do its work. What do you think @shinh, @ukai?

At any rate, it should at least be 0664, like it was.

pphaneuf: Other flags have a period at the end of their description string, also.

theopolis: Good catch on the default, o666 sounds good.

"some" flags have a period. If you prefer a period just ask :)

pphaneuf: I'd still like the opinion of @shinh or @ukai on the default mode (I'm really just a "helper" maintainer), but this seems good to me.

My personal preference for flag description string is "not capitalized, no period", because they're in parentheses within an existing "sentence", like this:

```
    -stop_logging_if_full_disk (Stop attempting to log to disk if the disk is
      full.) type: bool default: false
```

I only looked at the flags immediately surrounding it, but since it's already kind of random, I'll have to live with this (for now), thanks! :wink:

theopolis: Haha! I'm just messing around :dancer: ...but the gflags default help _could_ use some love.

PR #30, opened 2015-05-20
support for msvs 2015 RC

KjellSchubert:
glog won't compile without these changes for msvs 2015 RC. I found the need for deleting the move ctor odd, especially since the libglog_static project compiled without it, only the libglog project needed it (apparently C++11 features cannot be disabled in msvs, see http://stackoverflow.com/questions/23386472/disable-c11-features-in-vs2013)

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -31,7 +31,6 @@
 
 #include "utilities.h"
 
-#include <algorithm>
 #include <assert.h>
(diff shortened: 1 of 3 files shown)


KjellSchubert:
I signed it

sergiud:
Changes seems to be obsolete. Please reopen if needed.

PR #42, opened 2015-07-26
Add windows command line build script.

haosdent:


sergiud:
I'm not sure this is needed anymore since we have a proper CMake support. Please reopen if you have a different opinion.

Issue #41, opened 2015-07-24
visual studio 2015 error 


ml232528:
error C2084: 函数“int snprintf(char *const ,const size_t,const char *const ,...)”已有主体
error C2280: “std::basic_ios<char,std::char_traits<char>>::basic_ios(const std::basic_ios<char,std::char_traits<char>> &)”: 尝试引用已删除的函数

ml232528:
modify logging.h
class GOOGLE_GLOG_DLL_DECL LogStream : public std::ostream {
...
  public:
	  LogStream(const LogStream &) = delete;
	  void operator=(const LogStream &) = delete; 
 };

modify port.cc
#if !(defined(_MSC_VER) && (_MSC_VER >= 1900))
int snprintf(char *str, size_t size, const char *format, ...) {
  va_list ap;
  va_start(ap, format);
  const int r = vsnprintf(str, size, format, ap);
  va_end(ap);
  return r;
}
#endif

ukai:
which is better?  #43 and #30 
do we need both?

sergiud:
Neither. 856ff81a8268a5c22d026a65d4c12a2e1136f73f and f9def39a465725c1b2a60f64200fba5d1614da36 already fixed VS2015 issues.

ukai:
ok. fixed by #7

PR #43, opened 2015-07-29
Port glog to Visual Studio 2015 RTM

hausdorff:
glog v0.3.4 has two outstanding issues with the code that prevent it
from compiling on the RTM of Visual Studio 2015 toolchain (and, more
generally, MSVC).

The first is that the function `snprintf` is defined in the version of
stdio.h that ships with the VS2015 toolchain. The fix for this is
relatively straightforward -- we simply define the function only for
versions before the 2015 version.

The second is that the 2015 toolchain will implicitly define a move
constructor for `LogMessage::LogStream`, which in turn will at some
point call a deleted copy constructor for basic_ios. Our solution in
this commit will be to simply explicitly delete it.

### DIFF
--- a/src/windows/port.cc
+++ b/src/windows/port.cc
@@ -55,6 +55,8 @@ int safe_vsnprintf(char *str, size_t size, const char *format, va_list ap) {
   return _vsnprintf(str, size-1, format, ap);
 }
 
+#if !(defined(_MSC_VER) && (_MSC_VER >= 1900))
+// MSVC 1900 defines this method body as part of stdio.h.
 int snprintf(char *str, size_t size, const char *format, ...) {
   va_list ap;
   va_start(ap, format);
@@ -62,3 +64,4 @@ int snprintf(char *str, size_t size, const char *format, ...) {
   va_end(ap);
   return r;
 }
+#endif
--- a/src/glog/logging.h.in
+++ b/src/glog/logging.h.in
@@ -1157,6 +1157,12 @@ public:
     int ctr() const { return ctr_; }
     void set_ctr(int ctr) { ctr_ = ctr; }
     LogStream* self() const { return self_; }
+#if defined(_MSC_VER) && (_MSC_VER >= 1900)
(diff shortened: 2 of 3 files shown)


hausdorff:
I signed an individual contributor agreement.

hausdorff:
This also resolves issue #41.

hausdorff:
Hate to bother you all, but is there any hope that this will be integrated into upstream? We need it for the [Apache Mesos](http://mesos.apache.org/) project.

haosdent:
+1

src/windows/port.cc
```diff
@@ -55,6 +55,8 @@ int safe_vsnprintf(char *str, size_t size, const char *format, va_list ap) {
   return _vsnprintf(str, size-1, format, ap);
 }
 
+#if !(defined(_MSC_VER) && (_MSC_VER >= 1900))
+// MSVC 1900 defines this method body as part of stdio.h.
```
mayah: I believe it's better to remove the corresponding declaration from port.h?

hausdorff: We still need it for MSVC versions 1800 and below. Only 1900 has implemented it.

mayah: I mean, it is better to write the same #if in port.h

mayah:
Anyway, we need this fix to build our product on VS2015, @ukai-san?

sergiud:
The newly merged CMake support PR includes all the changes required to compile glog using Microsoft Visual Studio 2015. Hence, this PR is redundant.

hausdorff:
Excellent. Delighted to hear. That is a better outcome than I could have reasonable expected. (I certainly am not the best person to port this code. :))

Sent from Outlook




On Thu, Aug 6, 2015 at 4:44 AM -0700, "Sergiu Dotenco" <notifications@github.com> wrote:












The newly merged CMake support PR includes all the changes required to compile glog using Microsoft Visual Studio 2015. Hence, this PR is redundant.

sergiud:
Please use CMake to generate Visual Studio solutions.

PR #51, opened 2015-08-19
Link libpthread when necessary in CMake builds

flandr:
When building on Linux, we need to pass -lpthread when using threads.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -376,6 +376,10 @@ if (UNWIND_LIBRARY)
   target_link_libraries (glog PUBLIC ${UNWIND_LIBRARY})
 endif (UNWIND_LIBRARY)
 
+if (HAVE_PTHREAD)
+  target_link_libraries (glog PUBLIC ${CMAKE_THREAD_LIBS_INIT})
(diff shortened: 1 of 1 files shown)


sergiud:
That not the correct way of fixing this. ``find_package (Threads)`` already locates pthreads. However, a call to ``target_link_libraries (glog PUBLIC ${CMAKE_THREAD_LIBS_INIT})`` is missing. Also, the line should be added only, if both ``WITH_THREADS`` and ``Threads_FOUND`` are true.

flandr:
SGTM, will fix

sergiud:
Great, thanks!

flandr:
See updated patch set. The `HAVE_PTHREAD` variable is set only when `WITH_THREADS && `Threads_FOUND`.

sergiud:
Looks good.

trygveu:
Thanks @flandr, I was just facing the linker problem here on my side. The works fine.

flandr:
Should we merge this, then?

sergiud:
Definitely. @ukai can do it.

PR #58, opened 2015-09-08
Fix missing public include directory

ruslo:
Path `<prefix>/include` will not be added to the headers search of dependent
targets without specifying INSTALL_INTERFACE, i.e. compilation
will fail with error:

  "fatal error: 'glog/logging.h' file not found".


ruslo:
I signed it!

sergiud:
:thumbsup:

PR #74, opened 2015-11-26
Reset SIGABRT action only if FailureSignalHandler is installed.

yoshisatoyanagisawa:
When I set my own signal handler to SIGABRT, it did not executed
with CHECK.  That is because SIGABRT handler is reset to default
just before glog calls abort.
Let me make it reset only if the handler is what glog installed
i.e. FailureSignalHandler.


yoshisatoyanagisawa:
I am Googler.  I should have already signed CLA.

PR #76, opened 2015-12-03
android specific fixes

amir-sabbaghi:


amir-sabbaghi:
I signed it!

src/logging.cc
```diff
@@ -1110,7 +1110,7 @@ void LogFileObject::Write(bool force_flush,
        (bytes_since_flush_ >= 1000000) ||
        (CycleClock_Now() >= next_flush_time_) ) {
     FlushUnlocked();
-#ifdef OS_LINUX
+#if defined(OS_LINUX) && !defined(ANDROID)
```
ukai: newer SDK has posix_fadvise?
http://stackoverflow.com/questions/29347531/undefined-reference-to-posix-fadvise

src/logging_unittest.cc
```diff
@@ -637,7 +637,7 @@ static void GetFiles(const string& pattern, vector<string>* files) {
   } while (FindNextFileA(handle, &data));
   BOOL result = FindClose(handle);
   LOG_SYSRESULT(result);
-#else
+#elif !defined(ANDROID)
```
ukai: it could fix compile error, but would fail in test?
I think CheckFille would fail at CHECK_EQ(files.size(), 1UL).

src/raw_logging.cc
```diff
@@ -59,7 +59,7 @@
 # include <unistd.h>
 #endif
 
-#if defined(HAVE_SYSCALL_H) || defined(HAVE_SYS_SYSCALL_H)
+#if !defined(ANDROID) && (defined(HAVE_SYSCALL_H) || defined(HAVE_SYS_SYSCALL_H))
```
ukai: why !defined(ANDROID)?
could you explain the reason?

src/logging.cc
amir-sabbaghi: I do not know. I am compiling for android 4.x which does not have posix_fadvice. And it's a little soon to only support Android 5.x, as there are less than 25% Android 5.x users.

src/logging_unittest.cc
amir-sabbaghi: Android does not have glob.h

src/raw_logging.cc
amir-sabbaghi: Android does not have SYS_write.

amir-sabbaghi: andoird-1 has a header "sys/glibc-syscalls.h" which contains "SYS_write" definition, but it has been deprecated a long time ago.
newer platforms does not have this header.

src/logging.cc
shinh: I think you can use __ANDROID_API__?

src/raw_logging.cc
shinh: I think you can use __NR_write. I think Binoic doesn't consider __NR_* are public APIs and the code may not compile in future, but I think we really don't want raw_logging to modify errno.

amir-sabbaghi: I really think we should not use private APIs.
It's safer to modify errno instead of using private API.

sergiud:
Changes seems to be obsolete. Please reopen if needed.

PR #81, opened 2015-12-30
Resolve missing prototype warning

Nazg-Gul:
Makes compilation process less noisy even when using string compiler flags.


Nazg-Gul:
Needed to re-arrange some branches in own clone, re-submitted this pull-request as #302.

PR #82, opened 2016-01-08
Better support and documentation for CMake projects that consume glog

sergiud:
CMake adds an imported target with the namespaced alias `glog::glog`. Also, added initial CMake usage documentation.


sergiud:
@ukai Could you please merge this PR?

PR #88, opened 2016-03-04
Respect configured docdir

kekstee:
Avoid overriding the configured docdir inside the Makefile.
$(datarootdir)/doc/ may differ from $(prefix)/share/doc/


kekstee:
Signed.

PR #96, opened 2016-03-31
Update src/windows/glog/logging.h to ToT

mayah:
src/windows/glog/logging.h is not generated by the current source.
Let me run src/windows/preprocess.sh and update the source.


mayah:
I signed it!

PR #110, opened 2016-05-10
Fix compile error for clang++ on Linux

rcgoodfellow:
Hi,

I was unable to compile glog using clang++ on Linux due to GOOGLE_GLOG_DLL_DECL being treated as a symbol instead of a macro in demangle.h. This patch fixes the problem as simply as possible, I considered handling it at the configure.ac level, but figured that should be a decision for you guys.

Thanks!
~ ry

### DIFF
--- a/src/demangle.h
+++ b/src/demangle.h
@@ -72,6 +72,10 @@
 
 #include "config.h"
 
+#if defined(__linux__)
+  #define GOOGLE_GLOG_DLL_DECL
+#endif
+
 _START_GOOGLE_NAMESPACE_
 
 // Demangle "mangled".  On success, return true and write the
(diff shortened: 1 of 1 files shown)


rcgoodfellow:
I signed it!

rcgoodfellow:
Also, [here](https://gist.github.com/rcgoodfellow/5e9d385d9813c742312b6e86dbf6e1e4) is my build script for glog. The `autoreconf -i` is because I am using Ubuntu 16.04 which has a newer version of ac-local than the configure script in source is looking for.

sergiud:
This is not a clang specific issue. #116 fixes the problem.

PR #116, opened 2016-06-23
Fix autotools build.

pcc:
It looks like commit 3c49b93 modified the auto-generated file src/config.h.in
to add a definition of macro GOOGLE_GLOG_DLL_DECL. One of the autotools
reverts this change upon running "make", causing the build to fail when a
source file includes demangle.h.

To fix the problem, revert the change to src/config.h.in and include
glog/logging.h from demangle.h which provides a definition of that macro.


sergiud:
@ukai Please merge.

PR #123, opened 2016-09-15
Set sinks_ to NULL after deletion in LogDestination::DeleteLogDestinations

yying:
`LogDestination::DeleteLogDestinations` deletes `sinks_` but does not set the global to `NULL`. 

Further calls to `LogDestination::AddLogSink` will incorrectly use the stale `sinks_` pointer and potentially crash (https://github.com/google/glog/blob/master/src/logging.cc#L621).

This simple fix sets `sinks_` to `NULL` after the `sinks_` is deleted.


yying:
I signed it!

PR #148, opened 2016-12-30
support ISO-8601 timestamp format

leeyiw:
Some monitor system don't recognize glog's timestamp format. And currently glog don't print year. This MR add support for ISO-8601 timestamp format, and add an option: `FLAGS_iso8601format = 1;` which is default set to `false`.

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -115,6 +115,8 @@ GLOG_DEFINE_bool(alsologtostderr, BoolFromEnv("GOOGLE_ALSOLOGTOSTDERR", false),
                  "log messages go to stderr in addition to logfiles");
 GLOG_DEFINE_bool(colorlogtostderr, false,
                  "color messages logged to stderr (if supported by terminal)");
+GLOG_DEFINE_bool(iso8601format, BoolFromEnv("GOOGLE_ISO8601FORMAT", false),
+                 "format messages' timestamp using ISO 8601 standard");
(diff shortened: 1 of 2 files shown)


leeyiw:
I signed it!

leeyiw:
@ukai Could this PR be merged?

src/logging.cc
```diff
@@ -1381,15 +1383,28 @@ void LogMessage::Init(const char* file,
...
+      stream() << setw(2) << 1900+data_->tm_time_.tm_year << "-"
+               << setw(2) << 1+data_->tm_time_.tm_mon << "-"
+               << setw(2) << data_->tm_time_.tm_mday
+               << "T"
+               << setw(2) << data_->tm_time_.tm_hour << ':'
+               << setw(2) << data_->tm_time_.tm_min << ':'
+               << setw(2) << data_->tm_time_.tm_sec
+               << (data_->tm_time_.tm_gmtoff >= 0 ? "+" : "-")
+               << setw(2) << ::labs(data_->tm_time_.tm_gmtoff)/(60*60) << ":"
+               << setw(2) << ::labs(data_->tm_time_.tm_gmtoff)%(60*60);
```
sergiud: `tm_gmtoff` is not available for MSVC and MinGW.

leeyiw: Is there any portable solution for `tm_gmtoff`?

PR #151, opened 2017-01-04
Port stack tracing to Windows

andschwa:
This set of patches implements the stack trace, symbolizer, and demangler for Windows, and their unit tests. It utilizes the Windows debugging APIs in `"Dbghelp.h"`. I'd like this to start being reviewed, but feel it probably needs some more unit tests and definitely needs to be integration tested. I'll also clean up the commits before it's applied. This _is_ a work in progress, with more up-to-date work on my [test branch](https://github.com/andschwa/glog/tree/test). I mostly need to fix how the stack trace reporting is called, as using the signal handler does not seem to result in the desired behavior on Windows.

/cc @hausdorff

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -455,6 +455,11 @@ if (HAVE_EXECINFO_H)
   set (HAVE_STACKTRACE 1)
 endif (HAVE_EXECINFO_H)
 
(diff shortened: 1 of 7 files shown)


andschwa:
I signed it!

PR #153, opened 2017-01-10
cmake: do not hardcode relative install directories

sergiud:
The relative install directories are determined using the CMake `GNUInstallDirs` module. In particular, this allows to automatically determine the correct `lib` directory (`lib` vs `lib64` etc.).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,7 @@ include (CMakePackageConfigHelpers)
 include (CPack)
 include (CTest)
 include (DetermineGflagsNamespace)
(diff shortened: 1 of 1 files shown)


sergiud:
@shinh Please merge.

andschwa:
Just curious, does this work as expected (or is at least compatible with) Windows?

andschwa:
(I'm guessing yes, because I doubt anyone uses the `install` target on Windows.)

sergiud:
It does. On Windows the commit won't make any difference. Under Linux, the correct suffix for `lib` will be detected automatically.

PR #158, opened 2017-01-18
[RFC] reduce heap memory allocations to zero

sergiud:
In case there's thread-local storage (TLS) support, logging doesn't need to allocate memory for every message making it more memory efficient.

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -336,6 +336,7 @@ const size_t LogMessage::kMaxLogMessageLen = 30000;
 
 struct LogMessage::LogMessageData  {
   LogMessageData();
+  void reset();
 
   int preserved_errno_;      // preserved errno
   // Buffer space; contains complete message text.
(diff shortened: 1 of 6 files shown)


hansres:
I confirm being okay with my commits being contributed to this project.

sergiud:
@ukai Could you please take a look?

src/logging.cc
```diff
@@ -1198,8 +1206,19 @@ void LogMessage::Init(const char* file,
                       void (LogMessage::*send_method)()) {
   allocated_ = NULL;
   if (severity != GLOG_FATAL || !exit_on_dfatal) {
+#ifdef GLOG_THREAD_LOCAL_STORAGE
+    // No need for locking, because the data is thread local.
+    if (thread_data_available) {
+      thread_data_available = false;
+      data_ = new (&thread_msg_data) LogMessageData();
+    } else {
+      allocated_ = new LogMessageData();
```
shinh: I'd add a test case for this code path. Just

LOG(INFO) << LOG(INFO) << "nested LOG";

should be fine.

sergiud:
@shinh Done. Please review.

PR #159, opened 2017-01-22
Sanitize Autotools

SoapGentoo:
Hi @ukai 
I've cleaned up the build system a bit to match modern Autotools standards.

* Remove all Autotools-generated cruft from the repo and add it to `.gitignore`.
* Generate `.bz2` tarballs instead of `.gz` tarballs.
* Make "Google Tests", "Google Flags" and "Google Test" have a corresponding `--disable` flag, which is less automagic than currently.
* Do not override `$(docdir)` in `Makefile.am`, as this is a user flag in the GNU conventions.
* Make tests only compile when invoked with `make check`.


SoapGentoo:
@sergiud any chance of getting this merged?

sergiud:
Obsoleted by #585.

PR #167, opened 2017-02-22
Avoid creating large amounts of `-WUnused` in projects using glog

ChrisKitching:
`logging.h` contains a fancy macro that emulates `static_assert` via an interesting use of `typedef`. In situations where the assertion is true, the typedef ends up unused, which yields a warning on some recent compilers.

This PR redefines the macro to use the C++ built-in `static_assert` when compiling for C++11 or newer (since that is when the built-in was added). This avoids the warnings, and provides slightly nicer errors when the assert fails.

This seems a more elegant solution to the problem than attempting to directly suppress the warning, since there seems to be no non-insane way to do that in a compiler-agnostic way.

### DIFF
--- a/src/glog/logging.h.in
+++ b/src/glog/logging.h.in
@@ -934,8 +934,13 @@ struct CrashReason;
 bool IsFailureSignalHandlerInstalled();
 }  // namespace glog_internal_namespace_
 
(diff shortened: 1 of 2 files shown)


jray272:
Fixed in PR #255

jray272:
Err PR #225

sergiud:
Closing since another PR has been merged.

PR #168, opened 2017-03-04
Port stack tracing to Windows

andschwa:
This set of patches implements the stack trace, symbolizer, and demangler for Windows, and their unit tests. It utilizes the Windows debugging APIs in `"Dbghelp.h"`.

There is one `TODO` for `IsFailureSignalHandlerInstalled` as I'm unsure yet how to implement it for Windows. That said, in order to make stack trace reporting work "as expected" for test failures, the data dump needs to be done immediately, instead of in the signal handler (see commit message for details).

I've integration tested this with Mesos, and so far it is working well. I've also updated the unit tests where appropriate.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -455,6 +455,11 @@ if (HAVE_EXECINFO_H)
   set (HAVE_STACKTRACE 1)
 endif (HAVE_EXECINFO_H)
 
+if (WIN32)
+  set (HAVE_STACKTRACE 1)
+  target_link_libraries (glog PUBLIC Dbghelp.lib)
+endif (WIN32)
+
 if (UNIX OR (APPLE AND HAVE_DLADDR))
   set (HAVE_SYMBOLIZE 1)
 endif (UNIX OR (APPLE AND HAVE_DLADDR))
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -364,9 +364,9 @@ set (GLOG_SRCS
   src/vlog_is_on.cc
 )
 
-if (HAVE_PTHREAD)
+if (HAVE_PTHREAD OR WIN32)
   list (APPEND GLOG_SRCS src/signalhandler.cc)
-endif (HAVE_PTHREAD)
+endif (HAVE_PTHREAD OR WIN32)
 
 if (WIN32)
   list (APPEND GLOG_SRCS
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -457,6 +457,7 @@ endif (HAVE_EXECINFO_H)
 
 if (WIN32)
   set (HAVE_STACKTRACE 1)
+  set (HAVE_SYMBOLIZE 1)
   target_link_libraries (glog PUBLIC Dbghelp.lib)
 endif (WIN32)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -458,7 +458,6 @@ endif (HAVE_EXECINFO_H)
 if (WIN32)
   set (HAVE_STACKTRACE 1)
   set (HAVE_SYMBOLIZE 1)
-  target_link_libraries (glog PUBLIC Dbghelp.lib)
 endif (WIN32)
 
 if (UNIX OR (APPLE AND HAVE_DLADDR))
--- a/src/demangle.cc
+++ b/src/demangle.cc
@@ -35,10 +35,16 @@
 // Note that we only have partial C++0x support yet.
 
 #include <stdio.h>  // for NULL
+#include "utilities.h"
 #include "demangle.h"
+
+#if defined(OS_WINDOWS)
+#include <DbgHelp.h>
+#endif
 
 _START_GOOGLE_NAMESPACE_
 
+#if !defined(OS_WINDOWS)
 typedef struct {
   const char *abbrev;
   const char *real_name;
@@ -1293,12 +1299,37 @@ static bool ParseTopLevelMangledName(State *state) {
   }
   return false;
 }
+#endif
 
 // The demangler entry point.
 bool Demangle(const char *mangled, char *out, int out_size) {
+#if defined(OS_WINDOWS)
+  // When built with incremental linking, the Windows debugger
+  // library provides a more complicated `Symbol->Name` with the
+  // Incremental Linking Table offset, which looks like
+  // `@ILT+1105(?func@Foo@@SAXH@Z)`. However, the demangler expects
+  // only the mangled symbol, `?func@Foo@@SAXH@Z`. Fortunately, the
+  // mangled symbol is guaranteed not to have parentheses,
+  // so we search for `(` and extract up to `)`.
+  //
+  // Since we may be in a signal handler here, we cannot use `std::string`.
+  char buffer[1024];  // Big enough for a sane symbol.
+  const char *lparen = strchr(mangled, '(');
+  if (lparen) {
+    // Extract the string `(?...)`
+    const char *rparen = strchr(lparen, ')');
+    size_t length = rparen - lparen - 1;
+    strncpy(buffer, lparen + 1, length);
+    buffer[length] = '\0';
+    mangled = buffer;
+  } // Else the symbol wasn't inside a set of parentheses
+  // We use the ANSI version to ensure the string type is always `char *`.
+  return UnDecorateSymbolName(mangled, out, out_size, UNDNAME_COMPLETE);
+#else
   State state;
   InitState(&state, mangled, out, out_size);
   return ParseTopLevelMangledName(&state) && !state.overflowed;
+#endif
 }
 
 _END_GOOGLE_NAMESPACE_
--- a/src/demangle.cc
+++ b/src/demangle.cc
@@ -40,6 +40,7 @@
 
 #if defined(OS_WINDOWS)
 #include <DbgHelp.h>
+#pragma comment(lib, "DbgHelp")
 #endif
 
 _START_GOOGLE_NAMESPACE_
--- a/src/signalhandler.cc
+++ b/src/signalhandler.cc
@@ -48,9 +48,6 @@
 
 _START_GOOGLE_NAMESPACE_
 
-// TOOD(hamaji): Use signal instead of sigaction?
-#ifdef HAVE_SIGACTION
-
 namespace {
 
 // We'll install the failure signal handler for these signals.  We could
@@ -66,9 +63,13 @@ const struct {
   { SIGILL, "SIGILL" },
   { SIGFPE, "SIGFPE" },
   { SIGABRT, "SIGABRT" },
+#if !defined(OS_WINDOWS)
   { SIGBUS, "SIGBUS" },
+#endif
   { SIGTERM, "SIGTERM" },
 };
+
+static bool kFailureSignalHandlerInstalled = false;
 
 // Returns the program counter from signal context, NULL if unknown.
 void* GetPC(void* ucontext_in_void) {
@@ -168,6 +169,9 @@ void DumpTimeInfo() {
   g_failure_writer(buf, formatter.num_bytes_written());
 }
 
+// TOOD(hamaji): Use signal instead of sigaction?
+#ifdef HAVE_SIGACTION
+
 // Dumps information about the signal to STDERR.
 void DumpSignalInfo(int signal_number, siginfo_t *siginfo) {
   // Get the signal name.
@@ -213,6 +217,8 @@ void DumpSignalInfo(int signal_number, siginfo_t *siginfo) {
   g_failure_writer(buf, formatter.num_bytes_written());
 }
 
+#endif  // HAVE_SIGACTION
+
 // Dumps information about the stack frame to STDERR.
 void DumpStackFrameInfo(const char* prefix, void* pc) {
   // Get the symbol name.
@@ -240,12 +246,17 @@ void DumpStackFrameInfo(const char* prefix, void* pc) {
 
 // Invoke the default signal handler.
 void InvokeDefaultSignalHandler(int signal_number) {
+#ifdef HAVE_SIGACTION
   struct sigaction sig_action;
   memset(&sig_action, 0, sizeof(sig_action));
   sigemptyset(&sig_action.sa_mask);
   sig_action.sa_handler = SIG_DFL;
   sigaction(signal_number, &sig_action, NULL);
   kill(getpid(), signal_number);
+#elif defined(OS_WINDOWS)
+  signal(signal_number, SIG_DFL);
+  raise(signal_number);
+#endif
 }
 
 // This variable is used for protecting FailureSignalHandler() from
@@ -256,9 +267,14 @@ static pthread_t* g_entered_thread_id_pointer = NULL;
 
 // Dumps signal and stack frame information, and invokes the default
 // signal handler once our job is done.
+#if defined(OS_WINDOWS)
+void FailureSignalHandler(int signal_number)
+#else
 void FailureSignalHandler(int signal_number,
                           siginfo_t *signal_info,
-                          void *ucontext) {
+                          void *ucontext)
+#endif
+{
   // First check if we've already entered the function.  We use an atomic
   // compare and swap operation for platforms that support it.  For other
   // platforms, we use a naive method that could lead to a subtle race.
@@ -298,16 +314,20 @@ void FailureSignalHandler(int signal_number,
   // First dump time info.
   DumpTimeInfo();
 
+#if !defined(OS_WINDOWS)
   // Get the program counter from ucontext.
   void *pc = GetPC(ucontext);
   DumpStackFrameInfo("PC: ", pc);
+#endif
 
 #ifdef HAVE_STACKTRACE
   // Get the stack traces.
   void *stack[32];
   // +1 to exclude this function.
   const int depth = GetStackTrace(stack, ARRAYSIZE(stack), 1);
+# ifdef HAVE_SIGACTION
   DumpSignalInfo(signal_number, signal_info);
+# endif
   // Dump the stack traces.
   for (int i = 0; i < depth; ++i) {
     DumpStackFrameInfo("    ", stack[i]);
@@ -333,18 +353,19 @@ void FailureSignalHandler(int signal_number,
 
 }  // namespace
 
-#endif  // HAVE_SIGACTION
-
 namespace glog_internal_namespace_ {
 
 bool IsFailureSignalHandlerInstalled() {
 #ifdef HAVE_SIGACTION
+  // TODO(andschwa): Return kFailureSignalHandlerInstalled?
   struct sigaction sig_action;
   memset(&sig_action, 0, sizeof(sig_action));
   sigemptyset(&sig_action.sa_mask);
   sigaction(SIGABRT, NULL, &sig_action);
   if (sig_action.sa_sigaction == &FailureSignalHandler)
     return true;
+#elif defined(OS_WINDOWS)
+  return kFailureSignalHandlerInstalled;
 #endif  // HAVE_SIGACTION
   return false;
 }
@@ -363,11 +384,18 @@ void InstallFailureSignalHandler() {
   for (size_t i = 0; i < ARRAYSIZE(kFailureSignals); ++i) {
     CHECK_ERR(sigaction(kFailureSignals[i].number, &sig_action, NULL));
   }
+  kFailureSignalHandlerInstalled = true;
+#elif defined(OS_WINDOWS)
+  for (size_t i = 0; i < ARRAYSIZE(kFailureSignals); ++i) {
+    CHECK_NE(signal(kFailureSignals[i].number, &FailureSignalHandler),
+             SIG_ERR);
+  }
+  kFailureSignalHandlerInstalled = true;
 #endif  // HAVE_SIGACTION
 }
 
 void InstallFailureWriter(void (*writer)(const char* data, int size)) {
-#ifdef HAVE_SIGACTION
+#if defined(HAVE_SIGACTION) || defined(OS_WINDOWS)
   g_failure_writer = writer;
 #endif  // HAVE_SIGACTION
 }
--- a/src/stacktrace_windows-inl.h
+++ b/src/stacktrace_windows-inl.h
@@ -1,0 +1,59 @@
+// Copyright (c) 2000 - 2007, Google Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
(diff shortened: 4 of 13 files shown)


andschwa:
Also note: I updated the `CMakeLists.txt` build system, as that's what I'm familiar with, but there appear to be several Windows build systems maintained in parallel.

kaysoky:
Note: The contents of this PR have been applied to the Windows build of Apache Mesos (https://github.com/apache/mesos/commit/ef3ebf3fda475ac0d8d51d12ba5c2e10cb0607df).

andschwa:
@ukai can a maintainer please take a look?

andschwa:
@ukai and @shinh anything blocking this, please, can it at least be reviewed?

andschwa:
Adding @sergiud for a possible review.

andschwa:
Ping @sergiud @ukai @shinh

src/signalhandler.cc
```diff
@@ -145,7 +145,7 @@ void DumpTimeInfo() {
 void DumpStackFrameInfo(const char* prefix, void* pc) {
   // Get the symbol name.
   const char *symbol = "(unknown)";
-  char symbolized[1024];  // Big enough for a sane symbol.
+  char symbolized[1024] = {};  // Big enough for a sane symbol.
```
shinh: I couldn't understand why this change is necessary. When Symbolize returns true, SymbolizeAndDemangle fills something to this buffer on both Windows and non-Windows?

src/symbolize.cc
```diff
@@ -837,6 +837,57 @@ static ATTRIBUTE_NOINLINE bool SymbolizeAndDemangle(void *pc, char *out,
...
+  }
+  // Resolve symbol information from address.
+  // https://msdn.microsoft.com/en-us/library/windows/desktop/ms680578(v=vs.85).aspx
+  char buf[sizeof(SYMBOL_INFO) + MAX_SYM_NAME * sizeof(TCHAR)];
+  SYMBOL_INFO *symbol = reinterpret_cast<SYMBOL_INFO*>(buf);
+  symbol->SizeOfStruct = sizeof(SYMBOL_INFO);
+  symbol->MaxNameLen = MAX_SYM_NAME;
+  bool ret = SymFromAddr(symInitializer.process, reinterpret_cast<DWORD64>(pc), 0, symbol);
+  if (ret && static_cast<int>(symbol->NameLen) < out_size) {
+      strncpy(out, symbol->Name, symbol->NameLen + 1);
```
shinh: Two space indentation please.

src/symbolize.cc
```diff
@@ -837,6 +837,57 @@ static ATTRIBUTE_NOINLINE bool SymbolizeAndDemangle(void *pc, char *out,
...
+  if (!symInitializer.ready) {
+    return false;
+  }
+  // Resolve symbol information from address.
+  // https://msdn.microsoft.com/en-us/library/windows/desktop/ms680578(v=vs.85).aspx
+  char buf[sizeof(SYMBOL_INFO) + MAX_SYM_NAME * sizeof(TCHAR)];
+  SYMBOL_INFO *symbol = reinterpret_cast<SYMBOL_INFO*>(buf);
+  symbol->SizeOfStruct = sizeof(SYMBOL_INFO);
+  symbol->MaxNameLen = MAX_SYM_NAME;
+  bool ret = SymFromAddr(symInitializer.process, reinterpret_cast<DWORD64>(pc), 0, symbol);
```
shinh: Could you break after the first comma? We usually don't use more than 80 columns per a line.

src/symbolize_unittest.cc
```diff
@@ -353,18 +386,15 @@ int main(int argc, char **argv) {
...
   printf("PASS (no symbolize_unittest support)\n");
+  return 0;
+# endif  // __ELF__
 #else
   printf("PASS (no symbolize support)\n");
-#endif
   return 0;
+#endif
 }
-#endif  // HAVE_STACKTRACE
```
shinh: Could you add // HAVE_SYMBOLIZE? (apparently the previous comment was wrong..)

src/stacktrace_windows-inl.h
```diff
@@ -27,33 +27,60 @@
...
-  if (result_count < 0)
-    result_count = 0;
-  if (result_count > max_depth)
-    result_count = max_depth;
-  for (int i = 0; i < result_count; i++)
-    result[i] = stack[i + skip_count];
+  if (max_depth > 64) {
+    max_depth = 64;
+  }
+  std::vector<void*> stack(max_depth);
```
shinh: Note: On ELF systems, Symbolize and Demangle are meant to be async-signal-safe. As signal handler can be called even while malloc's internal lock is held, I think we use no malloc in Symbolize and Demangle. I don't know Windows signal handler. I assume it's OK to call any functions?

Hmm it seems to be no? https://msdn.microsoft.com/en-us/library/aa272921(v=vs.60).aspx

I'm also worried by Windows API calls to obtain stacktrace/symbol. I guess some of them are calling malloc internally.

src/signalhandler.cc
```diff
@@ -363,11 +392,15 @@ void InstallFailureSignalHandler() {
   for (size_t i = 0; i < ARRAYSIZE(kFailureSignals); ++i) {
     CHECK_ERR(sigaction(kFailureSignals[i].number, &sig_action, NULL));
   }
+#elif defined(OS_WINDOWS)
+  typedef void (*SignalHandlerPointer)(int); 
+  SignalHandlerPointer previousHandler;  
+  previousHandler = signal(SIGABRT, &FailureSignalHandler);
```
shinh: Are you sure you only care SIGABRT? I think handling SIGFPE, SIGILL, and SIGSEGV would be useful.

src/demangle.cc
```diff
@@ -1293,12 +1299,32 @@ static bool ParseTopLevelMangledName(State *state) {
...
 bool Demangle(const char *mangled, char *out, int out_size) {
+#if defined(OS_WINDOWS)
+  // When built with incremental linking, the Windows debugger
+  // library provides a more complicated `Symbol->Name` with the
+  // Incremental Linking Table offset, which looks like
+  // `@ILT+1105(?func@Foo@@SAXH@Z)`. However, the demangler expects
+  // only the mangled symbol, `?func@Foo@@SAXH@Z`. Fortunately, the
+  // mangled symbol is guaranteed not to have parentheses,
+  // so we search for `(` and extract up to `)`.
+  std::string symbol(mangled);
```
shinh: Again, it'd be better to avoid malloc.

shinh: Hmm I'm confused. I think I've left a comment in stacktrace_windows-inl.h or somewhere, but the comment has disappeared :(

Anyway, I think calling malloc in signal handler isn't safe, so it'd be nicer to avoid using std::string. https://msdn.microsoft.com/en-us/library/aa272921(v=vs.60).aspx

I'm also worried by Windows APIs.

src/signalhandler.cc
```diff
@@ -168,6 +141,76 @@ void DumpTimeInfo() {
   g_failure_writer(buf, formatter.num_bytes_written());
 }
 
+// Dumps information about the stack frame to STDERR.
+void DumpStackFrameInfo(const char* prefix, void* pc) {
+  // Get the symbol name.
+  const char *symbol = "(unknown)";
+  char symbolized[1024] = {};  // Big enough for a sane symbol.
```
shinh: I couldn't understand why this change is necessary. When Symbolize returns true, SymbolizeAndDemangle fills something to this buffer on both Windows and non-Windows?

shinh:
Thanks for the patch! Sorry for the huge latency... Could you update AUTHORS and CONTRIBUTORS as well (see CONTRIBUTING.md)? I'm a bit worried by the use of malloc and Windows APIs in signal handler, but I think it's definitely much nicer than nothing. So I'm planning to merge this PR if others don't objects.

CMakeLists.txt
```diff
@@ -455,6 +455,12 @@ if (HAVE_EXECINFO_H)
   set (HAVE_STACKTRACE 1)
 endif (HAVE_EXECINFO_H)
 
+if (WIN32)
+  set (HAVE_STACKTRACE 1)
+  set (HAVE_SYMBOLIZE 1)
+  target_link_libraries (glog PUBLIC Dbghelp.lib)
```
sergiud: I would remove the `.lib` suffix here and replace `Dbghelp` by proper capitalization of library's name, i.e., `DbgHelp `.

src/demangle.cc
```diff
@@ -35,10 +35,16 @@
 // Note that we only have partial C++0x support yet.
 
 #include <stdio.h>  // for NULL
+#include "utilities.h"
 #include "demangle.h"
 
+#if defined(OS_WINDOWS)
+#include <Dbghelp.h>
```
sergiud: For consistency reasons, same as before, i.e., `DbgHelp.h` instead of `Dbghelp.h`. Windows Dev Center also suggests this header name.

src/symbolize.cc
```diff
@@ -837,6 +837,57 @@ static ATTRIBUTE_NOINLINE bool SymbolizeAndDemangle(void *pc, char *out,
 
 _END_GOOGLE_NAMESPACE_
 
+#elif defined(OS_WINDOWS)
+
+#include "Dbghelp.h"
```
sergiud: Same as before.

src/symbolize.cc
```diff
@@ -837,6 +837,57 @@ static ATTRIBUTE_NOINLINE bool SymbolizeAndDemangle(void *pc, char *out,
...
+
+_START_GOOGLE_NAMESPACE_
+
+class SymInitializer {
```
sergiud: This class should not be copied, correct? To ensure no one does this in the future, I'd delete the copy constructor and copy assignment operator by adding the declarations
```c++
SymInitializer(const SymInitializer&);
SymInitializer& operator=(const SymInitializer&);
```
to the private section of the class.

andschwa:
@serguid and @shinh Thank you so much for getting to this and reviewing it! I'll commit to updating it per reviews (plus feedback on some confusing Windows-y parts) by the end of the week. Talk to you soon!

src/signalhandler.cc
andschwa: Perhaps this was not the case on Linux, but on Windows, the `Symbolize()` only filled the buffer as far as the symbol took. So you'd run into this kind of problem:
```
[uninitializedbufferofrandomdata]
[main()ializedbufferofrandomdata]
[someLongSymbol()ferofrandomdata]
[short()gSymbol()ferofrandomdata]
```
Since the buffer was not being initialized, but the same data location was being reused and only partially filled out, previous longer symbols left gunk in buffer with new shorter symbols.

I solved this by ensuring all elements were initialized to zero each time `DumpStackFrameInfo` was called. In retrospect, if I ensure a the terminating null is added manually, this becomes unnecessary.

CMakeLists.txt
andschwa: I agree, and even more so, I think I should instead use:

```
#pragma comment(lib, "DbgHelp")
```

In an appropriate header, eliminating the need for us to `target_link_libraries` or for consumers of Glog to do so either.

src/symbolize.cc
andschwa: Heh, thanks.

src/symbolize.cc
andschwa: Also thanks. Not sure how I messed that up.

src/demangle.cc
andschwa: Got it. This is where I'm going to add the `pragma` for automatic linking as well.

src/signalhandler.cc
andschwa: This is where Windows gets weird. While `SIGNAL.H` defines all of these signals as valid parameters to `signal()`, it does so only for compatibility. Explicitly, the signals `SIGILL` and `SIGTERM` are not generated on Windows (though you can explicitly raise them).

However, I think that since you can explicitly raise these signals (and it appears that `SIGFPE` and `SIGSEGV` might be raised on Windows), I'll change this to handle them all.

src/symbolize_unittest.cc
andschwa: This took me a little bit to figure out haha, fixing.

src/demangle.cc
andschwa: You're right, I need to fix this up.

andschwa: (I found your other comment by the way, it was getting hidden under "show outdated" even though I don't think the code was actually outdated. Not sure what GitHub is up to.)

src/signalhandler.cc
andschwa: Also note:

I ran into a major problem on Windows with signal handling. Let me reproduce the commit message here, as the commit is going to change when I update the PR:

> On Windows, the signal handler is invoked on the thread on which it was installed, which is not necessarily the thread where the test failed. So instead of using the signal handler to dump the stack trace, which would dump it from the wrong thread, we dump the information immediately.

> Doing this causes us to use a few more frames, so `DumpStackInfo` needed to configurable skip number.

> The nicer approach would be to refactor the `MinimalFormatter` and info dumping functions out of the signal handler and into the logging library, which is a more relevant home for them, but it is less invasive to simply move the two functions Windows needed into the internal namespace.

So the _usual_ code path that a test failure takes now skips signal handler entirely, and goes straight to dumping stack and symbols.

andschwa: I _think_ this is because the Windows signal handler is not like what is found on Linux like systems. Instead, it uses structured exception handling (SEH). I don't know much more than what I've researched on this (ironically I am not a Windows dev, I am a Linux dev), but I know this was quite an annoying problem to solve. Signaling on Windows switches threads (to the SEH thread, I guess?) and `CaptureStackBackTrace` explicitly gets the backtrace for the current thread. I'm looking for a way to do this right.

src/signalhandler.cc
andschwa: I think this comment got duplicated.

src/signalhandler.cc
andschwa: Oh [here it is](https://github.com/google/glog/pull/168/files#diff-8b7c2e002f1814c322f81e0c0ecb3e7cR1481).

The signal handler was being on the thread on which it was installed, not a dedicated SEH thread (I was mis-remembering), and importantly, not the thread on which the signal was raised.

src/symbolize.cc
andschwa: Got it.

src/symbolize.cc
andschwa: Thanks.

andschwa:
With a fresh set of eyes (and more Windows knowledge) I actually managed to get the signal handler working properly today (just tested an integration with Mesos). I'll get it cleaned up and the PR posted.

src/signalhandler.cc
andschwa: I fixed this. I did a bit more investigating, and now this works as expected. You install the failure signal handlers, and thing work like they do on Linux.

andschwa:
@shinh and @sergiud this is ready for a final review I believe. I fixed the mentioned issues, and fixed the deeper issue of the signal handler not being all the way hooked up. From my previous comments, I mentioned that I'd plumbed around the signal handler "not working as expected" on Windows; but I now have it working as expected, and undid the unnecessary changes. Users should use `InstallFailureSignalHandler()` and expect `LOG(FATAL)`, `CHECK(...)` and `abort()` etc. to all work.

src/symbolize.cc
```diff
@@ -837,6 +837,66 @@ static ATTRIBUTE_NOINLINE bool SymbolizeAndDemangle(void *pc, char *out,
...
+    SymCleanup(process);
+  }
+private:
+  SymInitializer(const SymInitializer&);
+  SymInitializer& operator=(const SymInitializer&);
+};
+
+__declspec(noinline) static bool SymbolizeAndDemangle(void *pc, char *out,
+                                                      int out_size) {
+  const SymInitializer symInitializer;
```
shinh: I'd guess SymInitialize is a fairly complex function. Does it have no malloc?

Also, is it a good idea to construct SymInitializer every time? Perhaps the results of SymInitialize will be cached internally and the second SymInitialize call is fast?

andschwa: I think I meant this to be a `const static` variable so it was initialized only once.

andschwa: Fixed.

andschwa: > I'd guess SymInitialize is a fairly complex function. Does it have no malloc?

I'm unsure. I suppose we could construct `SymInitialize` at startup (i.e. move `symInitializer` out of the function here).

andschwa:
Just to add some test results:

CentOS 7 (ensuring I didn't break anything):
```sh
$ ./autogen.sh
$ ./configure
$ make -j8
$ make check
============================================================================
Testsuite summary for glog 0.3.5
============================================================================
# TOTAL: 6
# PASS:  6
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================

$ mkdir build
$ cd build
$ cmake3 ..
$ make -j8
$ ctest
Test project /home/andschwa/src/glog/build
    Start 1: demangle
1/6 Test #1: demangle .........................   Passed    0.00 sec
    Start 2: logging
2/6 Test #2: logging ..........................   Passed    0.53 sec
    Start 3: signalhandler
3/6 Test #3: signalhandler ....................   Passed    0.01 sec
    Start 4: stacktrace
4/6 Test #4: stacktrace .......................   Passed    0.00 sec
    Start 5: stl_logging
5/6 Test #5: stl_logging ......................   Passed    0.00 sec
    Start 6: symbolize
6/6 Test #6: symbolize ........................   Passed    0.00 sec

100% tests passed, 0 tests failed out of 6

Total Test time (real) =   0.55 sec
```

Windows 10:
```powershell
$ mkdir build
$ cd build
$ cmake ..
$ cmake --build . --config Debug
$ ctest -C Debug
Test project C:/Users/andschwa/src/glog/build
    Start 1: demangle
1/6 Test #1: demangle .........................   Passed    0.05 sec
    Start 2: logging
2/6 Test #2: logging ..........................***Failed    0.47 sec
    Start 3: signalhandler
3/6 Test #3: signalhandler ....................   Passed    0.02 sec
    Start 4: stacktrace
4/6 Test #4: stacktrace .......................***Failed    0.97 sec
    Start 5: stl_logging
5/6 Test #5: stl_logging ......................   Passed    0.01 sec
    Start 6: symbolize
6/6 Test #6: symbolize ........................   Passed    0.03 sec

67% tests passed, 2 tests failed out of 6

Total Test time (real) =   1.58 sec

The following tests FAILED:
          2 - logging (Failed)
          4 - stacktrace (Failed)
Errors while running CTest
```

The stacktrace test tends to fail in `Debug` because of the incremental linking table making function and return addresses unreliable. The symbolizer should now be expected to fail because the `Release` configuration doesn't have symbol information.

```powershell
$ cmake --build . --config Release
$ ctest -C Release
Test project C:/Users/andschwa/src/glog/build
    Start 1: demangle
1/6 Test #1: demangle .........................   Passed    0.00 sec
    Start 2: logging
2/6 Test #2: logging ..........................***Failed    0.45 sec
    Start 3: signalhandler
3/6 Test #3: signalhandler ....................   Passed    0.02 sec
    Start 4: stacktrace
4/6 Test #4: stacktrace .......................   Passed    0.02 sec
    Start 5: stl_logging
5/6 Test #5: stl_logging ......................   Passed    0.01 sec
    Start 6: symbolize
6/6 Test #6: symbolize ........................***Exception: Other  3.15 sec

67% tests passed, 2 tests failed out of 6

Total Test time (real) =   3.69 sec

The following tests FAILED:
          2 - logging (Failed)
          6 - symbolize (OTHER_FAULT)
Errors while running CTest
```

And finally, the logging test fails just due to a difference in padding:
```
$ ctest -V -C Debug
2: Test with golden file failed. We'll try to show the diff:
2: --- C:/Users/andschwa/src/glog/src/logging_unittest.err.munged       Tue Jun 27 11:36:39 2017
2: +++ C:\Users\andschwa\AppData\Local\Temp\/captured.err.munged        Tue Jun 27 11:36:39 2017
2: @@ -78,7 +78,7 @@
2:  IDATE TIME__ THREADID logging_unittest.cc:LINE] RAW: foo bar 10 3.400000
2:  WDATE TIME__ THREADID logging_unittest.cc:LINE] RAW: array
2:  IDATE TIME__ THREADID logging_unittest.cc:LINE] RAW: const array
2: -IDATE TIME__ THREADID logging_unittest.cc:LINE] RAW: ptr 12345678
2: +IDATE TIME__ THREADID logging_unittest.cc:LINE] RAW: ptr 0000000012345678
2:  IDATE TIME__ THREADID logging_unittest.cc:LINE] RAW: ptr 0000000000000000
2:  EDATE TIME__ THREADID logging_unittest.cc:LINE] RAW: foo 1000 0000001000 3e8
2:  IDATE TIME__ THREADID logging_unittest.cc:LINE] RAW: foo 1000
2: diff command was failed.
2: Check failed: MungeAndDiffTestStderr(FLAGS_test_srcdir + "/src/logging_unittest.err")
2/6 Test #2: logging ..........................***Failed    0.45 sec
```
The golden file has `12345678` and we print `0000000012345678`.

Finally, I've also linked this library into Mesos and tested some expected failures with an `abort()`, a `CHECK_EQ(true, false)` and a `LOG(FATAL) << Failing`, and stack traces are being generated properly.

src/symbolize.cc
shinh: > I'm unsure. I suppose we could construct SymInitialize at startup (i.e. move symInitializer out of the function here).

Yeah, but MSDN told me SymInitialize could be slow. Perhaps it might be a good idea to add a function which initializes SymInitialize if a library user wants to be safe. Anyway, it's up to you. I'm OK with the current patch. Probably I'll merge this PR tomorrow unless you want me to wait a bit more.

andschwa: @shinh that's where I'm at with it; I initialized it in the function since it can be slow. While I think adding a function to pre-initialize it for safety is a good idea, it's also not necessary for most use cases (judging by how I'm using this in Mesos anyways), so I'd rather not add it now.

I am good with the current contents of this PR for you to merge them. Thank you for your help!

shinh:
Yeah, further changes can be done in a separate PR if necessary. Thanks!

andschwa:
Woo! Thank you :)

src/utilities.cc
```diff
@@ -137,17 +137,19 @@ static void DumpStackTraceAndExit() {
...
     // own signal handler installed by InstallFailureSignalHandler().
+#ifdef HAVE_SIGACTION
     struct sigaction sig_action;
     memset(&sig_action, 0, sizeof(sig_action));
     sigemptyset(&sig_action.sa_mask);
     sig_action.sa_handler = SIG_DFL;
     sigaction(SIGABRT, &sig_action, NULL);
-  }
+#elif defined(OS_WINDOWS)
+    signal(SIGABRT, SIG_DFL);
```
andschwa: Oops, this should set `kFailureSignalHandlerInstalled` back to `false`.

bmahler:
Filed to issues to improve this: [issue 534](https://github.com/google/glog/issues/534) and [issue 535](https://github.com/google/glog/issues/535).

PR #179, opened 2017-04-20
CHECK_NOTNULL works with smart pointers when compiled in C++11.

renning22:
Tested on ubuntu by gcc4.8.4 with/without -std=c++11. The corresponding version of CheckNotNull was selected correctly and results were expected.


renning22:
I signed it!

PR #188, opened 2017-05-12
Remove files generated by autotools

shinh:
for #165. I think not many projects have them in their repo.


ArturKlajnerok:
Hi @ukai, any chance we could have a new release with those changes ? The `test-driver` symlink is causing some issues in https://github.com/facebook/react-native

PR #199, opened 2017-06-22
Use gflags ALIAS instead of ${gflags_XXX} vars

### LINKED ISSUE #198: Use gflags ALIAS target instead of ${gflags_XXX} variables
The gflags project recently added an ALIAS library target for `gflags` which can be used instead of the legacy `${gflags_INCLUDES}` and `${gflags_LIBRARIES}` variables. This will also allow removal of the `target_include_directories` line for gflags. Happy to submit a pull request if this sounds good.

v1bri:
The gflags project updated their CMake config last year with a `gflags` ALIAS target. This can be used instead of the legacy `${gflags_LIBRARIES}` and `${gflags_INCLUDE_DIRS}` variables. It also looks cleaner.

Fixes #198

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,7 +49,7 @@ include (DetermineGflagsNamespace)
 set (CMAKE_THREAD_PREFER_PTHREAD 1)
 
 if (WITH_GFLAGS)
-  find_package (gflags)
+  find_package (gflags 2.2.0)
 
   if (gflags_FOUND)
     set (HAVE_LIB_GFLAGS 1)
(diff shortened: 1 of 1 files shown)


sergiud:
Which gflags introduced the alias? This needs to be set [here](https://github.com/google/glog/blob/master/CMakeLists.txt#L52).

v1bri:
Updated with gflags version.

v1bri:
Any additional comments @sergiud?

sergiud:
@v1bri Looks good to me now.

v1bri:
Thanks! Would you like to merge @shinh or @ukai?

PR #200, opened 2017-06-24
Replace BUILD_TESTING with GLOG_BUILD_TESTS

pwnall:
This makes it easier to use glog as a subdirectory in a larger CMake project. This technique is borrowed from googletest.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,7 @@ set (CPACK_PACKAGE_VERSION ${GLOG_VERSION})
 
 option (WITH_GFLAGS "Use gflags" ON)
 option (WITH_THREADS "Enable multithreading support" ON)
+option (GLOG_BUILD_TESTS "Build all of glog's own tests." "${BUILD_TESTING}")
 
 list (APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
@@ -461,7 +462,7 @@ endif (UNIX OR (APPLE AND HAVE_DLADDR))
 
 # Unit testing
 
-if (BUILD_TESTING)
+if (GLOG_BUILD_TESTS)
   add_executable (logging_unittest
     src/logging_unittest.cc
   )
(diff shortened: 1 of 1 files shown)


sergiud:
Why can't you implement `GLOG_BUILD_TESTS` in your root CMake project by setting `BUILD_TESTING` before calling `add_subdirectory` and then simply restore `BUILD_TESTING`?

`BUILD_TESTING` is a standard CTest variable. Renaming it would cause several problems. Specifically, you disregarded the case, where one intentionally wants to disable all the tests using a single variable.

pwnall:
Thank you very much for the quick response!

1) I tried your suggestion (and stackoverflow suggested that it'd work), but for some reason CMake is using the restored value. I figured I'm probably doing something wrong, but I also saw that googletest declares its own variable instead of using `BUILD_TESTING`, and I know that googletest was design to be `add_subdirectory`ed in CMake builds, so I figured there must be a good reason to go down that route.

2) I tried addressing the concern you mentioned above by having `GLOG_BUILD_TESTS` default to the value of `BUILD_TESTING`. As long as the user doesn't manually tweak `GLOG_BUILD_TESTS`, setting `BUILD_TESTING` to off should still disable glog's tests.

sergiud:
How exactly did you set `BUILD_TESTING`?

pwnall:
```cmake
  set (BUILD_TESTING_SAVED "${BUILD_TESTING}")
  set (BUILD_TESTING OFF)
  add_subdirectory ("${PROJECT_SOURCE_DIR}/third_party/glog" EXCLUDE_FROM_ALL)
  set (BUILD_TESTING "${BUILD_TESTING_SAVED}")
```

Commenting out the last line does result in glog's tests not being built, and that makes me think that it's somehow picking up the restored value.

sergiud:
`BUILD_TESTING` is cache variable. Therefore, you must use the `set (BUILD_TESTING ... CACHE BOOL "" FORCE")` form of the set command: https://cmake.org/cmake/help/v3.8/command/set.html#set-cache-entry.

pwnall:
Thanks for the tip! The following seems to work.

```cmake
  set (BUILD_TESTING_SAVED "${BUILD_TESTING}")
  set (BUILD_TESTING OFF CACHE BOOL "" FORCE)
  add_subdirectory ("${PROJECT_SOURCE_DIR}/third_party/glog" EXCLUDE_FROM_ALL)
  set (BUILD_TESTING "${BUILD_TESTING_SAVED}" CACHE BOOL "" FORCE)
```

It's still a tad ugly, though 😄

sergiud:
I'm closing this PR since there's a clean workaround available.

PR #203, opened 2017-06-29
Enable ThreadSanitizer

toddlipcon:
This adds a cmake option WITH_TSAN which enables thread sanitizer
instrumentation for glog. Additionally, it imports dynamic_annotations.h
from gperftools and adds annotations for benign races.

This prevents TSAN false positives around vlog levels and the rawlog
time variables.


CMakeLists.txt
```diff
@@ -377,6 +381,10 @@ endif (WIN32)
 
 add_compile_options ($<$<BOOL:${HAVE_NO_UNNAMED_TYPE_TEMPLATE_ARGS}>:-Wno-unnamed-type-template-args>)
 
+if (WITH_TSAN)
+  add_compile_options (-fsanitize=thread -DDYNAMIC_ANNOTATIONS_ENABLED)
```
sergiud: Wouldn't be `if (WITH_TSAN AND HAVE_VALGRIND_H)` correct here? As it is now, users may enable TSAN for platforms/compilers which do not support it (e.g., MSVC). Ideally, one would also check if the compiler actually supports the `-fsanitze` flag first.

sergiud:
Closing due to missing CLA.

PR #204, opened 2017-06-30
added absolute cmake path to sources

nidegen:
I added the CMake List file path as prefix to the included files so it works when configured as subproject!:)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@ set (CPACK_PACKAGE_VERSION ${GLOG_VERSION})
 option (WITH_GFLAGS "Use gflags" ON)
 option (WITH_THREADS "Enable multithreading support" ON)
 
-list (APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+list (APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake)
 
 include (CheckCSourceCompiles)
 include (CheckCXXCompilerFlag)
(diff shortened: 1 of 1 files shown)


nidegen:
i signed it!

sergiud:
Why is this needed? Glog can already be included using `add_subdirectory` as is.

nidegen:
ok, makes sense. i was adding it with 'include'. probably the wrong way:)

nidegen:
@sergiud actually, `add_subdirectory` only works when it actually is a subdirectory. in my case, it is not, hence it would come in handy...

i am trying to include it in an iOS target. maybe there is an easier way to build and link it to an iOS project?

nidegen:
as i deleted the branch already, i will open another PR

sergiud:
I cannot follow what you are doing. Can you elaborate? Why doesn't ` add_subdirectory` work? What's the error? Can you provide the CMake output?

nidegen:
it was
`
CMake Error at /Users/nidegen/Developer/OpenMapper/OpenMapper/CMakeLists.txt:59 (ADD_SUBDIRECTOR
Y):
  add_subdirectory not given a binary directory but the given source
  directory "/Users/nidegen/Developer/OpenMapper/OpenMapper/thirdparty/glog"
  is not a subdirectory of
  "/Users/nidegen/Developer/OpenMapper/OpenMapper/wrapper/ios".  When
  specifying an out-of-tree source a binary directory must be explicitly
  specified.
`
i guess i can add the binary dir. i am doing this as a side project, should look more carefully:)

nidegen:
however, it completely messes up the build folder structure that way..

nidegen:
the absolute paths would def make life easier

PR #205, opened 2017-07-05
Modify the CMakeList.txt

chaoer:
Add a switch (  `BUILD_SHARED_LIBS_SWITCH` ) to control will we build a shared libs. (Default is  `ON` ).

Anyway, it will be more convenient to control what types of libs will we build when work with CMake GUI.


chaoer:
I signed it!

chaoer:
@googlebot I signed it!

andschwa:
Leaving aside the rebase that includes my commits...

Why do you need a switch? Just pass `-DBUILD_SHARED_LIBS=ON` (or `OFF`) when running `cmake`. [CMake](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) uses this to determine how to build ambiguous `add_library()` commands.

sergiud:
The new option is absolutely redundant. As @andschwa pointed out, there's a standard CMake `BUILD_SHARED_LIBS` option which already controls whether static or shared libraries are compiled. As any other option, you can add it and control it through CMake GUI. No need for another option.

chaoer:
@andschwa @sergiud  
Yeah, I get it.
I am not so familiar with CMake, and I thought that `BUILD_SHARED_LIBS` is not a standard CMake option.
Anyway, I know it now, Thanks.

PR #208, opened 2017-07-09
Symbolizer support for mingw and cygwin

sergiud:
This PR fixes compilation errors with MinGW/MSYS2 and Cygwin while enabling support for both compiler platforms. The unit tests, however, do not pass yet. `TrySymbolize` always returns `NULL`. 

@andschwa Is there a chance you can look into this?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -454,10 +454,6 @@ if (WIN32)
   target_compile_definitions (glog PUBLIC GLOG_NO_ABBREVIATED_SEVERITIES)
 endif (WIN32)
 
-if (MSVC)
-  target_compile_definitions (glog PRIVATE _SCL_SECURE_NO_WARNINGS)
-endif (MSVC)
-
 set_target_properties (glog PROPERTIES PUBLIC_HEADER "${GLOG_PUBLIC_H}")
 
 set (_glog_CMake_BINDIR ${CMAKE_INSTALL_BINDIR})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,6 +112,7 @@ check_cxx_compiler_flag (-Wunnamed-type-template-args
 check_symbol_exists (snprintf stdio.h HAVE_SNPRINTF)
 
 check_library_exists (unwind get_static_proc_name "" HAVE_LIB_UNWIND)
+check_library_exists (DbgHelp UnDecorateSymbolName "" HAVE_DBGHELP)
 
 find_library (UNWIND_LIBRARY NAMES unwind DOC "unwind library")
 mark_as_advanced (UNWIND_LIBRARY)
@@ -339,6 +340,22 @@ else (HAVE_USING_OPERATOR)
   set (ac_cv_cxx_using_operator 0)
 endif (HAVE_USING_OPERATOR)
 
+if (HAVE_EXECINFO_H)
+  set (HAVE_STACKTRACE 1)
+endif (HAVE_EXECINFO_H)
+
+if (WIN32)
+  set (HAVE_STACKTRACE 1)
+
+  if (HAVE_DBGHELP)
+    set (HAVE_SYMBOLIZE 1)
+  endif (HAVE_DBGHELP)
+endif (WIN32)
+
+if (UNIX OR (APPLE AND HAVE_DLADDR))
+  set (HAVE_SYMBOLIZE 1)
+endif (UNIX OR (APPLE AND HAVE_DLADDR))
+
 set (SIZEOF_VOID_P ${CMAKE_SIZEOF_VOID_P})
 
 if (WITH_THREADS AND Threads_FOUND)
@@ -408,6 +425,10 @@ if (UNWIND_LIBRARY)
   target_link_libraries (glog PUBLIC ${UNWIND_LIBRARY})
 endif (UNWIND_LIBRARY)
 
+if (HAVE_DBGHELP)
+   target_link_libraries (glog PUBLIC DbgHelp)
+endif (HAVE_DBGHELP)
+
 if (HAVE_PTHREAD)
   target_link_libraries (glog PUBLIC ${CMAKE_THREAD_LIBS_INIT})
 endif (HAVE_PTHREAD)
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -414,6 +431,10 @@ if (UNWIND_LIBRARY)
   target_link_libraries (glog PUBLIC ${UNWIND_LIBRARY})
 endif (UNWIND_LIBRARY)
 
+if (HAVE_DBGHELP)
+   target_link_libraries (glog PUBLIC DbgHelp)
```
andschwa: At least with MSVC, this isn't necessary due to
```
#pragma comment(lib, "DbgHelp")
```
from [here](https://github.com/google/glog/blob/5cfa9a28f51c41bcafd993f6e93a28daf977d276/src/demangle.cc#L43). But I'm guessing it's needed for MinGW / Cygwin? Let me get one installed and give it a whirl.

CMakeLists.txt
```diff
@@ -439,6 +460,10 @@ if (WIN32)
   target_compile_definitions (glog PUBLIC GLOG_NO_ABBREVIATED_SEVERITIES)
 endif (WIN32)
 
+if (MSVC)
+  target_compile_definitions (glog PRIVATE _SCL_SECURE_NO_WARNINGS)
```
andschwa: I didn't particularly want to add this, as it's disabling valid warnings, where the code itself probably should be fixed (at some point). I chose not to hide them. Do you think we should do this, and if so, should we add at least a TODO to fix the source of the warnings?

src/symbolize.cc
```diff
@@ -837,18 +837,22 @@ static ATTRIBUTE_NOINLINE bool SymbolizeAndDemangle(void *pc, char *out,
...
 _END_GOOGLE_NAMESPACE_
 
-#elif defined(OS_WINDOWS)
+#elif defined(OS_WINDOWS) || defined(__CYGWIN__)
 
+#include <windows.h>
 #include <DbgHelp.h>
+
+#ifdef _MSC_VER
 #pragma comment(lib, "DbgHelp")
```
andschwa: Ah, yup, the above makes sense now.

src/symbolize.cc
```diff
@@ -837,18 +837,22 @@ static ATTRIBUTE_NOINLINE bool SymbolizeAndDemangle(void *pc, char *out,
 
 _END_GOOGLE_NAMESPACE_
 
-#elif defined(OS_WINDOWS)
+#elif defined(OS_WINDOWS) || defined(__CYGWIN__)
```
andschwa: I would suggest using `OS_CYGWIN` (from [here](https://github.com/google/glog/blob/5cfa9a28f51c41bcafd993f6e93a28daf977d276/src/utilities.h#L40)) for consistency.

src/symbolize.cc
```diff
@@ -837,18 +837,22 @@ static ATTRIBUTE_NOINLINE bool SymbolizeAndDemangle(void *pc, char *out,
...
 _START_GOOGLE_NAMESPACE_
 
 class SymInitializer {
 public:
-  HANDLE process = NULL;
-  bool ready = false;
-  SymInitializer() {
+  HANDLE process;
+  bool ready;
+  SymInitializer() : process(NULL), ready(false) {
```
andschwa: 👍

src/symbolize_unittest.cc
```diff
@@ -357,10 +357,12 @@ void ATTRIBUTE_NOINLINE TestWithReturnAddress() {
 #endif
 }
 
-# elif defined(OS_WINDOWS)
+# elif defined(OS_WINDOWS) || defined(__CYGWIN__)
```
andschwa: Ditto for `OS_CYGWIN`.

src/symbolize_unittest.cc
```diff
@@ -378,7 +380,13 @@ TEST(Symbolize, SymbolizeWithDemangling) {
 }
 
 __declspec(noinline) void TestWithReturnAddress() {
-  void *return_address = _ReturnAddress();
+  void *return_address =
+#ifdef __GNUC__ // Cygwin and MinGW support
+	  __builtin_return_address(0)
```
andschwa: 👍

src/symbolize_unittest.cc
```diff
@@ -401,7 +409,7 @@ int main(int argc, char **argv) {
   TestWithPCInsideNonInlineFunction();
   TestWithReturnAddress();
   return RUN_ALL_TESTS();
-# elif defined(OS_WINDOWS)
+# elif defined(OS_WINDOWS) || defined(__CYGWIN__)
```
andschwa: Ditto `OS_CYGWIN`.

src/utilities.h
```diff
@@ -127,6 +127,7 @@
 # define HAVE_STACKTRACE
 #endif
 
+#ifndef HAVE_SYMBOLIZE
 // defined by gcc
 #if defined(__ELF__) && defined(OS_LINUX)
 # define HAVE_SYMBOLIZE
```
andschwa: I kind of wish we could move this check into configuration time, but then we'd need to do it in a couple build systems. Oh well.

CMakeLists.txt
```diff
@@ -345,6 +346,22 @@ else (HAVE_USING_OPERATOR)
   set (ac_cv_cxx_using_operator 0)
 endif (HAVE_USING_OPERATOR)
 
+if (HAVE_EXECINFO_H)
+  set (HAVE_STACKTRACE 1)
+endif (HAVE_EXECINFO_H)
+
+if (WIN32)
+  set (HAVE_STACKTRACE 1)
```
andschwa: The Windows stacktrace support also comes from the `DbgHelp` library, maybe this should be moved under the `if (HAVE_DBGHELP)`.

CMakeLists.txt
andschwa: Yup saw below, that must be the case.

CMakeLists.txt
andschwa: As long as code is getting corrected, let's delete [this unnecessary include](https://github.com/google/glog/blob/5cfa9a28f51c41bcafd993f6e93a28daf977d276/src/stacktrace_windows-inl.h#L37) as `CaptureStackBackTrace` comes from the Windows [kernel](https://msdn.microsoft.com/en-us/library/windows/desktop/bb204633(v=vs.85).aspx).

andschwa: (My bad.)

andschwa:
I got MinGW installed and built this, and it does successfully build, but same problem as you, tests aren't passing. Taking a look...

sergiud:
I pushed an additional commit which fixes compilation problems for compilers without C99 support. Since I fixed several C++98/C++03 issues in the original PR, I thought I'd attach the commit to this PR as well.

sergiud:
@andschwa I finally managed to address the issues from your review.

I guess fixing stacktrace support for MinGW is not easy. For this reason, I suggest we merge this PR so we can have (partially) working Cygwin/MinGW builds even though the stacktrace unit test fail. Related to this, the stacktrace support for these distributions should probably be disabled by default.

andschwa:
🎉

Issue #213, opened 2017-07-22
MSVC 2013 build has been broken


mmp:
Building TOT with MSVC2013 gives the error:

```
(ClCompile target) -> 
  C:\pbrt-v3\src\ext\glog\src\logging.cc(1154): error C2483: 'thread_msg_data' : object with constructor or destructor cannot be declared 'thread' [c:\pbrt-v3\build\src\ext\glog\glog.vcxproj]
```

It looks like this is due to the changes introduced in 2df0ca34aa3000dadf76633ca700abf0bf50756d.

jasminezz:
@mmp I have the same error ,building with VS2013. Have you solved it ?

mmp:
I haven't spent any time working on a work-around/fix.

siniyake:
I am stuck in this problem too. Has anyone solved it so far?

shinh:
Could someone test if this works?

https://github.com/google/glog/pull/217

tangketan:
It seems just changing thread_msg_data to pointer type is enough, no need to change the CMakeLists.txt.

sergiud:
Please test PR #226.

sergiud:
Fixed in 2fe6508e536032b321067a541b6f0a3a54089bc3.

PR #217, opened 2017-07-26
Two fixes for zero allocation logging

shinh:
- Do not use non-POD type for TLS. This is for Issue #213.
- Use GNU's __thread if it's available. On my linux box, cmake
  wrongly thinks __attribute__((thread)), which is for Cygwin,
  is available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,17 +203,24 @@ int main() { }
 " HAVE_CXX11_TLS)
 
 check_cxx_source_compiles ("
+__thread int tls;
+int main() { }
+" HAVE_GNUC_TLS)
+
+check_cxx_source_compiles ("
 __attribute__((thread)) int tls;
 int main() { }
 " HAVE_CYGWIN_TLS)
 
 if (WITH_TLS)
-  if (HAVE_CYGWIN_TLS)
+  if (HAVE_CXX11_TLS)
+    set (GLOG_THREAD_LOCAL_STORAGE thread_local)
+  elseif (HAVE_GNUC_TLS)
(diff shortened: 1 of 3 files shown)


sergiud:
Your PR seems to leak memory. I have a working patch for the related issues which does not allocate dynamic memory. I however need to fix the tests.

shinh:
IMO not freeing heap-allocated memory is fine as long as it can be accessible via a global pointer. Tools like valgrind should categorize this as "still reachable" not a leak. I agree it'd be nicer we can get rid of it though. I'm looking forward to your patch, thanks!

$ cat still_reachable.cc
__thread int* global;
int main() {
  global = new int();
}
$ g++ still_reachable.cc
$ valgrind --leak-check=full ./a.out
...
==137829== HEAP SUMMARY:
==137829==     in use at exit: 4 bytes in 1 blocks
==137829==   total heap usage: 1 allocs, 0 frees, 4 bytes allocated
==137829==
==137829== LEAK SUMMARY:
==137829==    definitely lost: 0 bytes in 0 blocks
==137829==    indirectly lost: 0 bytes in 0 blocks
==137829==      possibly lost: 0 bytes in 0 blocks
==137829==    still reachable: 4 bytes in 1 blocks
==137829==         suppressed: 0 bytes in 0 blocks

DariuszOstolski:
What if your glog client application create and destroy 1000 threads (because it is supposed to work continuously for 2 years and is creating 3 threads per day for various tasks) ?  At some point of time your pull request will bring a trouble(sooner or later) and not freeing heap allocated memory in C++ is real issue for every application.

sergiud:
Added #226.

Issue #214, opened 2017-07-24
VS2008(win 10) build error


UnadXiao:
When i building the glog code gives the error：
```
1>..\..\src\logging.cc(1162) : error C2039: 'reset' : is not a member of 'google::LogMessage::LogStream'
1>d:\program files (x86)\glog-master\glog-master\src\symbolize.h(133) : error C2065: 'uint64_t' : undeclared identifier
1>d:\program files (x86)\glog-master\glog-master\src\symbolize.h(133) : error C2146: syntax error : missing ')' before identifier 'pc'
1>d:\program files (x86)\glog-master\glog-master\src\symbolize.h(133) : error C2513: 'int *' : no variable declared before '='
1>d:\program files (x86)\glog-master\glog-master\src\symbolize.h(137) : error C2059: syntax error : ')'
```

andschwa:
VS2008 doesn't have `uint64_t` (or really any of `<stdint.h>`). Can you try a recent compiler?

[This implies](https://stackoverflow.com/questions/12970293/why-microsoft-visual-studio-cannot-find-stdint-h) that at least VS2012 and up will work.

UnadXiao:
Thx
I tried the MSVC2015/2017 compiler, unfortunately, got the same error:
vs2015 target platform version 8.1 platform toolset visual studio 2015(v140)
vs2017 target platform version 10.0.15063.0 platform toolset visual studio 2017(v141)
Win10 Release x64
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	C2375	'snprintf': redefinition; different linkage	libglog	D:\Program Files (x86)\glog-master\glog-master\src\windows\port.h	115	
Error	C2039	'reset': is not a member of 'google::LogMessage::LogStream'	libglog	D:\Program Files (x86)\glog-master\glog-master\src\logging.cc	1162	
Error	C2375	'snprintf': redefinition; different linkage	libglog	D:\Program Files (x86)\glog-master\glog-master\src\windows\port.h	115	
Error	C2375	'snprintf': redefinition; different linkage	libglog	D:\Program Files (x86)\glog-master\glog-master\src\windows\port.h	115	
Error	C2375	'google::GetStackTrace': redefinition; different linkage	libglog	d:\program files (x86)\glog-master\glog-master\src\stacktrace_windows-inl.h	42	
Error	C2375	'snprintf': redefinition; different linkage	libglog	D:\Program Files (x86)\glog-master\glog-master\src\windows\port.h	115	
[...]

PR #218, opened 2017-07-26
Update Windows docs

andschwa:
This removes the nine year old Visual Studio Solution, and the updated readme explains why.

This should help avoid issues like #214.


shinh:
Nice, thanks!

PR #225, opened 2017-08-06
Fix LOG_EVERY_N with clang -Wunused-local-typedef

### LINKED ISSUE #223: LOG_EVERY_N warns with clang -Wunused-local-typedef
Any usage of LOG_EVERY_N fails to compile due to the compile assert used to check the severity parameter. This generates errors that look something like:
warning: unused typedef 'INVALID_REQUESTED_LOG_SEVERITY' [-Wunused-local-typedef].

This makes it impossible to use LOG_EVERY_N with this warning treated as an error.

You can see this occurring in the following links (although in those links it's just a warning):
https://stackoverflow.com/questions/37963886/errors-when-installing-caffe-on-mac-os-x-10-10-5
https://github.com/BVLC/caffe/issues/3608

I have a fix for this issue.

jray272:
Please note, this should wait until PR #224 has been submitted as that PR includes adding me to the contributors.

Glog uses a pre-C++11 compile time assert to verify the validity of
the severity parameter for LOG_EVERY_N. Unfortunately, some compilers
will complain about the usage of LOG_EVERY_N with
"-Wunused-local-typedef" due to the way the compile time assert is
constructed. This makes it impossible to use LOG_EVERY_N with this
warning treated as an error.

The fix simply removes the assert entirely. This is safe to do since
you can't put anything invalid into the severity parameters without
generating a compile error elsewhere. This has been safe to do ever
since the GLOG_ prefixes were added as part of 6febec361e.

Fixes #223

### DIFF
--- a/src/glog/logging.h.in
+++ b/src/glog/logging.h.in
@@ -934,9 +934,6 @@ struct CrashReason;
 bool IsFailureSignalHandlerInstalled();
 }  // namespace glog_internal_namespace_
 
-#define GOOGLE_GLOG_COMPILE_ASSERT(expr, msg) \
-  typedef @ac_google_namespace@::glog_internal_namespace_::CompileAssert<(bool(expr))> msg[bool(expr) ? 1 : -1]
-
 #define LOG_EVERY_N(severity, n)                                        \
   SOME_KIND_OF_LOG_EVERY_N(severity, (n), @ac_google_namespace@::LogMessage::SendToLog)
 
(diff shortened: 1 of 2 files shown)


shinh:
Sounds reasonable. I believe this is the only user of GOOGLE_GLOG_COMPILE_ASSERT. Could you remove this macro altogether?

jray272:
Will do.

jray272:
Removed in dd19fb2

shinh:
Thanks!

PR #226, opened 2017-08-08
Zero allocation fix

sergiud:
This is an alternative to PR #217 which does not allocate dynamic memory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,6 +193,11 @@ int main() { return i; }
 " HAVE_NAMESPACES)
 
 check_cxx_source_compiles ("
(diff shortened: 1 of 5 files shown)


sergiud:
@shinh The PR is ready to be merged!

shinh:
Looks great, thanks! I'm fairly sure LOG(INFO) is still allocation free, but have you somehow checked this didn't increase memory allocation?

sergiud:
I have no numbers. However, @hansres has been running the patch for an extended period of time and maybe he can comment on that.

Nevertheless, since there's just one static array per thread, the memory consumpution should not bet higher than without TLS.

PR #233, opened 2017-08-22
add explicit Dbghelp dependency on Windows for compilers other than MSVC

Optiligence:


andschwa:
I have no objection to this. It's redundant when using MSVC, but shouldn't cause a problem with it, and adds better CMake integration with non-MSVC compilers.

👍

CMakeLists.txt
```diff
@@ -410,6 +410,8 @@ add_library (glog
 
 set_target_properties (glog PROPERTIES POSITION_INDEPENDENT_CODE ON)
 
+target_link_libraries(${PROJECT_NAME} PUBLIC $<$<PLATFORM_ID:Windows>:Dbghelp>)
```
andschwa: `s/${PROJECT_NAME}/glog` for consistency with the other uses of the target.

Optiligence: done

sergiud:
I've addressed this in PR #208 already.

andschwa:
Oh yeah, I forgot about that. Thanks @sergiud.

@Optiligence you can close this, it's getting fixed. Thanks for your contribution.

Optiligence:
I like my line more^^ (`DbgHelp` should always be available anyway)
Nice to hear it’s getting fixed.

sergiud:
You patch is incomplete. The platform ID does not imply the existence of `Dbghelp`. You need to actually check whether the library exists.

Optiligence:
`Dbghelp` always exists, doesn’t it?

Optiligence:
For the functions being used it’s shipped with Windows since XP, i doubt further support is needed.

sergiud:
You're missing the point. Your check is too simplistic to cover all possibilities, e.g., in cross-compilation scenarios where dbghelp might actually not be available for whatever reasons.

Optiligence:
In general, I agree, but in this case I don’t. 
Similarly nobody is going to search for `user32` too. 
If `Dbghelp` isn’t available the code won’t compile anyway, so what’s the point.

PR #234, opened 2017-08-22
link to gflags_static when building static glog

Optiligence:


sergiud:
What's the rationale?

Optiligence:
I think it’s always built anyway, but I changed it to use COMPONENTS. 
Rationale is that I want to either build completely shared or completely static programs and therefore want all the components to either ship both and allow easy selection (e.g. via BUILD_SHARED_LIBS) or at least allow building either one without modifying their make file.

sergiud:
This breaks the existing behavior though. You cannot mix shared and static builds anymore.

Besides, what stops you from building a static and shared version of gflags and use it accordingly? Why does glog need to perform the selection?

Optiligence:
Although that’s true, I don’t have any reasonable rationale for such a mixed use case. 

For gflags the static library has a different name and therefore has to be handled explicitly.

Optiligence:
Welp, using the COMPONENTS approach doesn’t do what i want it to. 
If downstream projects add a `find_package(gflags COMPONENTS …)` call they “can” choose which gflags lib flavor to use, if no `find_package` is present CMake will just interpret `gflags` as a library name (therefore referring to the shared version, if found) instead of a target because gflags doesn’t use the colon notation which would prevent that.

sergiud:
You're not making any sense. What's a colon notation?

Optiligence:
`gflags::gflags` would be, which would error when that wasn’t a CMake target. 
Currently it’s possible that it is either a target (when a `find_package(gflags …` was done before) or it’s just interpreted as a library name.

sergiud:
I'll close this PR for now since the rationale for the breaking change is not clear.

PR #245, opened 2017-09-19
Fix username lookup in case of missing USER environment variable

DariuszOstolski:
This commit fixes an issue with invalid-user in file prefix when USER environment variable is not defined.

### DIFF
--- a/src/utilities.cc
+++ b/src/utilities.cc
@@ -47,6 +47,12 @@
 #ifdef HAVE_SYSLOG_H
 # include <syslog.h>
 #endif
+#ifdef HAVE_UNISTD_H
+# include <unistd.h>  // For geteuid.
+#endif
+#ifdef HAVE_PWD_H
+# include <pwd.h>
+#endif
 
 #include "base/googleinit.h"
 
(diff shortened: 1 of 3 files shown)


DariuszOstolski:
Ping?

src/utilities.cc
```diff
@@ -299,8 +305,26 @@ static void MyUserNameInitializer() {
   if (user != NULL) {
     g_my_user_name = user;
   } else {
-    g_my_user_name = "invalid-user";
+#if defined(HAVE_PWD_H) && defined(HAVE_UNISTD_H)
+    uid_t uid;
+    struct passwd pwd;
+    struct passwd*result = NULL;
```
shinh: Could you add a whitespace right before "result"?

struct passwd* result = NULL

src/utilities.cc
```diff
@@ -299,8 +305,26 @@ static void MyUserNameInitializer() {
...
+    char buffer[1024] = {'\0'};
+    uid = geteuid();
+    int pwuid_res = getpwuid_r(uid, &pwd, buffer, sizeof (buffer), &result);
```
shinh: Could you remove the whitespace between sizeof and buffer?

src/utilities.cc
```diff
@@ -299,8 +305,26 @@ static void MyUserNameInitializer() {
...
+    if(pwuid_res == 0) {
+      g_my_user_name = pwd.pw_name;
+    }
```
shinh: Please concatenate with the next line

src/utilities.cc
```diff
@@ -299,8 +305,26 @@ static void MyUserNameInitializer() {
...
   } else {
-    g_my_user_name = "invalid-user";
+#if defined(HAVE_PWD_H) && defined(HAVE_UNISTD_H)
+    uid_t uid;
+    struct passwd pwd;
+    struct passwd*result = NULL;
+    char buffer[1024] = {'\0'};
+    uid = geteuid();
+    int pwuid_res = getpwuid_r(uid, &pwd, buffer, sizeof (buffer), &result);
+    if(pwuid_res == 0) {
```
shinh: A whitespace after "if"

shinh:
Sorry for the latency..

DariuszOstolski:
Thanks

Issue #253, opened 2017-10-18
logging.h re-enables disabled warning C4275 (MSVC)


bjornpiltz:
The following source file
```c++
#include <QString>
#include <glog/logging.h>
#include <QException>
```
emits a warning, even though both Qt and glog explicitly disable that warning.
```
1>c:\qt\5.9.1\msvc2017_64\include\qtcore\qexception.h(60): warning C4275: non dll-interface class 'std::exception' used as base for dll-interface class 'QException'
1>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.11.25503\include\vcruntime_exception.h(43): note: see declaration of 'std::exception'
1>c:\qt\5.9.1\msvc2017_64\include\qtcore\qexception.h(59): note: see declaration of 'QException'
```

The following lines in logging.h are the culprits:
```c++
# pragma warning(disable: 4275)
  class GOOGLE_GLOG_DLL_DECL LogStream : public std::ostream {
# pragma warning(default: 4275)
```
They can be fixed as follows.
```c++
# pragma warning (push)
# pragma warning(disable: 4275)
  class GOOGLE_GLOG_DLL_DECL LogStream : public std::ostream {
# pragma warning (pop)
```
[There seems to be a consensus](https://stackoverflow.com/questions/4193476/is-using-pragma-warning-push-pop-the-right-way-to-temporarily-alter-warning-lev) that this is the right way to do it.

PR #254, opened 2017-10-18
#253: Use MS preprocessor idiom to disable warning

DariuszOstolski:
This PR solves a problem with re-enabled 4275 warning after including <glog/logging.h>


andschwa:
LGTM.

PR #261, opened 2017-11-03
Compute base addresses from program headers while reading /proc/self/…

pcc:
…maps.

We previously had logic to compute the base address from program
headers as part of symbolization. The problem is that we need a correct
base address earlier in order to adjust a PC into the image's address
space, as these addresses can appear in unsymbolized output.

There was previously an assumption that only the mapping that
was lowest in the address space did not need to be adjusted. This
assumption is not guaranteed (for example, the kernel may choose to
map an ET_DYN lowest) and in fact turned out to be wrong in binaries
linked with lld because the first mapping is read-only.

The solution is to move the program header reading logic into the
code that reads /proc/self/maps.

### DIFF
--- a/src/symbolize.cc
+++ b/src/symbolize.cc
@@ -56,6 +56,8 @@
 
 #if defined(HAVE_SYMBOLIZE)
 
+#include <string.h>
+
 #include <limits>
 
 #include "symbolize.h"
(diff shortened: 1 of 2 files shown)


pcc:
Addressed comments from Chromium review (https://chromium-review.googlesource.com/752753).

ukai:
could we have test for this?

I tried CC=/path/to/clang CXX=/path/to/clang++ ./configure && make && make check-TESTS
all tests passed without this CL.

src/symbolize.cc
```diff
@@ -590,6 +567,48 @@ OpenObjectFileContainingPcAndGetStartAddress(uint64_t pc,
       return -1;  // Malformed line.
     }
 
+    // Determine the base address by reading ELF headers in process memory.
+    ElfW(Ehdr) ehdr;
+    if (flags_start[0] == 'r' &&
+        ReadFromOffsetExact(mem_fd, &ehdr, sizeof(ElfW(Ehdr)), start_address) &&
+        memcmp(ehdr.e_ident, ELFMAG, SELFMAG) == 0) {
```
shinh: Could you add a comment for this? I think this is skipping non-text maps for usual cases.

shinh:
This PR and https://chromium-review.googlesource.com/c/chromium/src/+/752753 look great to me. I hope no project except Chromium is using InstallSymbolizeCallback. Quick google search seems to suggest there's no other code which uses InstallSymbolizeCallback. However, in case there're hidden users of this APIs, could you clearly mention this changes the semantics of InstallSymbolizeCallback? I'd like to send the message to glog's list after we merge this PR.

src/symbolize.cc
pcc: It is in fact skipping non-readable maps. I've added a comment.

pcc:
> could we have test for this?
>
> I tried CC=/path/to/clang CXX=/path/to/clang++ ./configure && make && make check-TESTS
all tests passed without this CL.

Done. The test makes sure that PIEs mapped at low addresses can be symbolized correctly.

> However, in case there're hidden users of this APIs, could you clearly mention this changes the semantics of InstallSymbolizeCallback?

I've added a note to the commit message.

ukai:
Thanks!
confirmed.

PR #264, opened 2017-11-09
add DCHECK_NEAR

ibtaylor:
I went to use this macro and noticed that it was missing. Seems like it should exist.

Tried to add a symmetric test for `CHECK_NEAR`, but nothing no such test appeared to exist, so I added a extremely simple test similar to existing tests.

I wasn't sure if I should update `src/windows/glog/logging.h` or if that was unintentionally in the repo to begin with.


sergiud:
Closing due to missing CLA.

PR #276, opened 2017-12-19
Add an example using glog from Bazel

drigz:
Combined with the coming job on ci.bazel.io, this should help keep the build working.

@iirina For the Bazel build.
@shinh For the example.

### DIFF
--- a/bazel/example/BUILD
+++ b/bazel/example/BUILD
@@ -1,0 +1,8 @@
+cc_test(
+    name = "main",
+    size = "small",
+    srcs = ["main.cc"],
+    deps = [
+        "//:glog",
+    ],
+)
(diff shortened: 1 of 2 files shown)


bazel/example/BUILD
```diff
@@ -0,0 +1,8 @@
+cc_test(
+    name = "main",
+    size = "small",
+    srcs = ["main.cc"],
+    deps = [
+        "@com_google_glog//:glog",
```
iirina: You don't need to reference the workspace name here since you are in the same workspace

WORKSPACE
```diff
@@ -1,5 +1,7 @@
+workspace(name = "com_google_glog")
```
iirina: This is also unnecessary because you don't have multiple workspaces

bazel/example/BUILD
drigz: Done. This was left over from my first attempt to have the example use com_google_glog as an external repo.

Thanks for the review!

WORKSPACE
drigz: Done.

PR #283, opened 2018-01-04
Use target gflags instead of old VARIABLES

Mizux:
As mentioned in [Use gflags ALIAS target instead of ${gflags_XXX} variables #198](https://github.com/google/glog/issues/198) do the same for the try_compile command...


Mizux:
This patch already belongs to Google LLC, it's just that github.com WYSIWYG use my personal address...

PR #291, opened 2018-01-31
Allow using glog/stl_logging.h with Bazel

### LINKED ISSUE #289: glog bazel rules seem to be broken
I've added glog in my Bazel WORKSPACE as follows:
```
git_repository(
    name = "com_github_glog_glog",
    commit = "3106945d8d3322e5cbd5658d482c9ffed2d892c0",
    remote = "https://github.com/google/glog.git",
)

bind(
    name = "glog",
    actual = "@com_github_glog_glog//:glog",
)
```

When I run 'bazel build //external:glog' it builds fine, however, if I try to add stl_logging.h as a dependency of a class, I get the following error:
```
bazel-out/darwin-fastbuild/bin/external/com_github_glog_glog/_virtual_includes/glog/glog/stl_logging.h:50:6: error: invalid token at start of a preprocessor expression
#if !@ac_cv_cxx_using_operator@
     ^
```

Looking at the generated file from the bazel rule under glog.bzl it looks like there is missing the string replacement to remove `ac_cv_cxx_using_operator`.

Next I tried to directly add the project under a third_party/glog/ directory in my repo to try to fix this locally in order to test, however, there seems to be an issue with the output of of the config_h rule not being found under `copts = [..., '-I%s/glog_internal' % gendir,...]`.

Thoughts on how to resolve either of these issues?

drigz:
Fixes #289.

### DIFF
--- a/bazel/example/main.cc
+++ b/bazel/example/main.cc
@@ -1,5 +1,6 @@
 #include <gflags/gflags.h>
 #include <glog/logging.h>
+#include <glog/stl_logging.h>
 
 int main(int argc, char* argv[]) {
   // Initialize Google's logging library.
@@ -9,6 +10,10 @@ int main(int argc, char* argv[]) {
   gflags::ParseCommandLineFlags(&argc, &argv, true);
 
   LOG(INFO) << "Hello, world!";
+
+  // glog/stl_logging.h allows logging STL containers.
+  std::vector<int> x {1, 2, 3};
+  LOG(INFO) << "ABC, it's easy as " << x;
 
   return 0;
 }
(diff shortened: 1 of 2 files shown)


bazel/example/main.cc
```diff
@@ -10,5 +11,9 @@ int main(int argc, char* argv[]) {
 
   LOG(INFO) << "Hello, world!";
 
+  // glog/stl_logging.h allows logging STL containers.
+  std::vector<int> x {1, 2, 3};
```
sergiud: This will not compile in pre C++11 mode. Is this intentional?

drigz: No, good point.

drigz:
Thanks for the review! PTAL?

sergiud:
@drigz LGTM!

PR #292, opened 2018-01-31
CMake Update

Mizux:
Few patch to:
- Ease the use of glog as subproject (i.e. adding alias glog::glog)
- Better manage gflags dependencies
- Rework VERSION management using project() provided variables
cf https://cmake.org/cmake/help/v3.0/command/project.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,21 +8,14 @@ if (POLICY CMP0063)
   cmake_policy (SET CMP0063 NEW)
 endif (POLICY CMP0063)
 
-project (glog)
-
-set (GLOG_MAJOR_VERSION 0)
-set (GLOG_MINOR_VERSION 3)
-set (GLOG_PATCH_VERSION 5)
-
-set (GLOG_VERSION
-  ${GLOG_MAJOR_VERSION}.${GLOG_MINOR_VERSION}.${GLOG_PATCH_VERSION})
+project(glog VERSION 0.3.5 LANGUAGES C CXX)
 
 set (CPACK_PACKAGE_NAME glog)
 set (CPACK_PACKAGE_DESCRIPTION_SUMMARY "")
-set (CPACK_PACKAGE_VERSION_MAJOR ${GLOG_MAJOR_VERSION})
-set (CPACK_PACKAGE_VERSION_MINOR ${GLOG_MINOR_VERSION})
-set (CPACK_PACKAGE_VERSION_PATCH ${GLOG_PATCH_VERSION})
-set (CPACK_PACKAGE_VERSION ${GLOG_VERSION})
+set (CPACK_PACKAGE_VERSION_MAJOR ${PROJECT_VERSION_MAJOR})
+set (CPACK_PACKAGE_VERSION_MINOR ${PROJECT_VERSION_MINOR})
+set (CPACK_PACKAGE_VERSION_PATCH ${PROJECT_VERSION_PATCH})
+set (CPACK_PACKAGE_VERSION ${PROJECT_VERSION})
 
 option (WITH_GFLAGS "Use gflags" ON)
 option (WITH_THREADS "Enable multithreading support" ON)
@@ -433,8 +426,8 @@ if (gflags_FOUND)
   endif (NOT BUILD_SHARED_LIBS)
 endif (gflags_FOUND)
 
-set_target_properties (glog PROPERTIES VERSION ${GLOG_MAJOR_VERSION})
-set_target_properties (glog PROPERTIES SOVERSION ${GLOG_VERSION})
+set_target_properties (glog PROPERTIES VERSION ${PROJECT_VERSION})
+set_target_properties (glog PROPERTIES SOVERSION ${PROJECT_VERSION_MAJOR})
 
 if (WIN32)
   target_compile_definitions (glog PUBLIC GLOG_NO_ABBREVIATED_SEVERITIES)
@@ -618,7 +611,7 @@ configure_package_config_file (glog-config.cmake.in
   NO_CHECK_REQUIRED_COMPONENTS_MACRO)
 
 write_basic_package_version_file (glog-config-version.cmake VERSION
-  ${GLOG_VERSION} COMPATIBILITY SameMajorVersion)
+  ${PROJECT_VERSION} COMPATIBILITY SameMajorVersion)
 
 export (TARGETS glog NAMESPACE glog:: FILE glog-targets.cmake)
 export (PACKAGE glog)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,8 +10,6 @@ endif (POLICY CMP0063)
 
 project (glog)
 
-enable_testing ()
-
 set (GLOG_MAJOR_VERSION 0)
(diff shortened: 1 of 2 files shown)


Mizux:
I signed it! (since I'm a Google Temp (red card))

CMakeLists.txt
```diff
@@ -53,8 +67,12 @@ set (CMAKE_THREAD_PREFER_PTHREAD 1)
 
 if (WITH_GFLAGS)
   find_package (gflags 2.2.0)
-
   if (gflags_FOUND)
+    if (TARGET gflags::gflags)
+      set(GFLAGS_TARGET gflags::gflags)
+    elseif(TARGET gflags)
+      set(GFLAGS_TARGET gflags)
```
andschwa: Just curious, what is this logic doing? Why is the target sometimes `gflags::gflags` and other times just `gflags`, is this a side effect of version differences for the `glags` package maybe? I'm not too familiar with it.

CMakeLists.txt
```diff
@@ -8,23 +31,14 @@ if (POLICY CMP0063)
   cmake_policy (SET CMP0063 NEW)
 endif (POLICY CMP0063)
 
-project (glog)
```
andschwa: I am 👍 on this, and checked that our [minimum CMake version](https://cmake.org/cmake/help/v3.0/command/project.html) supports this syntax in `project` (I think it must have been older 2.x versions that didn't).

CMakeLists.txt
sergiud: I guess, this is a workaround for gflags not providing an alias target when used as a subproject. 

@Mizux  Wouldn't it make sense to let the gflags folks add this alias instead of making this work using ugly hacks?

CMakeLists.txt
```diff
@@ -460,8 +453,8 @@ if (gflags_FOUND)
   endif (NOT BUILD_SHARED_LIBS)
 endif (gflags_FOUND)
 
-set_target_properties (glog PROPERTIES VERSION ${GLOG_MAJOR_VERSION})
-set_target_properties (glog PROPERTIES SOVERSION ${GLOG_VERSION})
+set_target_properties (glog PROPERTIES VERSION ${PROJECT_VERSION})
+set_target_properties (glog PROPERTIES SOVERSION ${PROJECT_VERSION_MAJOR})
```
sergiud: I'm not sure this is correct. Does this mean there will be only a `libglog.so.0` but not a `libglog.so.0.3.5`?

CMakeLists.txt
sergiud: I'm not. The policies must be set before a `project` call, not after. Please revert.

CMakeLists.txt
Mizux: Hi,
- Curently gflags-[target|config].cmake provide `gflags` imported target.
- There is no official findgflags.cmake module AFAIK
- I'm currently sending a [PR247 in gflags.git](https://github.com/gflags/gflags/pull/247) to use gflags::gflags instead (and providing alias as well when using as subproject)
- [gRPC](https://github.com/grpc/grpc/blob/master/cmake/gflags.cmake#L30) seems to already to look for gflags::gflags
- We have glog::glog protobuf::libprotobuf gRPC::grpc ...
- CMake preconise the use of namespace 
```
When providing imported targets, these should be namespaced (hence the Foo:: prefix); CMake will recognize that values passed to target_link_libraries() that contain :: in their name are supposed to be imported targets (rather than just library names), and will produce appropriate diagnostic messages if that target does not exist (see policy CMP0028).
```
src: https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html?highlight=namespace#a-sample-find-module (a ~bottom search for namespaced)

So all in all we have `gflags` currently but hope to have `gflags::gflags` soon -> here it is a workaround to be able to work during the transition since `find_package` don't provide a way to deal with this otherwise...

CMakeLists.txt
Mizux: No there will be 3 files: (-> means symbolic link aka ln -s ...)
libglog.so -> libglog.so.0 (since SO_VERSION which is the API compatibility is only major)
libglog.so.0 -> libglog.so.0.3.5 (so the "API" point on your build version)
libglog.so.0.3.5 (which is the "real" library)

As I said in the commit message, there is also an inversion between VERSION and SOVERSION usage
(easy to spot with make install DESTDIR=install && ls -l install/use/local/lib)
doc:  https://cmake.org/cmake/help/latest/prop_tgt/VERSION.html#prop_tgt:VERSION

Hope it's clear know :sweat_smile: 

side note: if you wrap your code in a python "setup.py" bdist will resolve the symlink so you just have to depend on `$<TARGET_SONAME_FILE:glog>`
cf https://github.com/google/or-tools/blob/mizux/dev/cmake/python.cmake#L110

CMakeLists.txt
Mizux: `project(VERSION)` came in CMake 3.0 (i.e. not available in 2.8.12)
Also Trusty cmake package is 2.8.12 BUT ubuntu provides 3.5.1 as `cmake3` package for trusty (and travis has trusty with 3.8 or 3.9 BTW).
So it should be reasonable to think everyone have at least CMake 3.0 with VERSION support...

andschwa: And we're using `cmake_minimum_required (VERSION 3.0)`, so it'll be a very explicit error if 2.x is used. It is not worth attempting to support CMake 2, it's possibly worse than than the Python 2 and 3 split IMHO.

CMakeLists.txt
sergiud: I would prefer not adding the alias hack and wait until the `gflags::gflags` alias is added. No need for temporary workarounds.

Mizux: LGTM remove the commit for this PR

Mizux:
I just remove the gflags::gflags / gflags management.
keep it on a dev branch on my fork when needed...

Mizux:
@googlebot rescan

CMakeLists.txt
```diff
@@ -1,3 +1,26 @@
...
+##   target_link_libraries(foo glog::glog)
+##
+## Example CMakeLists.txt of super-project which contains glog source tree:
+##   cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
+##   project(Foo)
+##
+##   add_subdirectory(glog)
+##
+##   add_executable(foo src/foo.cc)
+##   target_link_libraries(foo glog::glog)
```
sergiud: I believe, this is not a good place to document such specific stuff. I would suggest moving this part either into the wiki or `cmake/INSTALL.md`.

Mizux: Ok, I'll move it...

ps: I just put it here as [gflags CMakelists.txt](https://github.com/gflags/gflags/blob/master/CMakeLists.txt#L21) did

Mizux:
Ok I moved the doc INSTALL.md and incorparate this patch with the glog::glog alias commit.
You can see the rendering there https://github.com/Mizux/glog/blob/master/cmake/INSTALL.md

Don't hesitate to comment / modify / whatever...

sergiud:
LGTM.

PR #304, opened 2018-03-20
konsole family of terminfo supports colored output

Nazg-Gul:
Glog was missing colored output when running in terminal
who is set to konsole* TERM, even though the terminal
itself supports colored output.

Add extra terminfo to the check function, so now output
from Glog is properly colored.


shinh:
Thanks!

PR #305, opened 2018-03-21
Make int conversion explicit and fix clang warning

samuela:
Otherwise produces a warning: 
```
Implicit conversion loses integer precision: 'long' to 'int'
```


shinh:
Thanks!

PR #310, opened 2018-04-17
Fix __attribute__((noreturn)) mismatch warning in Bazel build

aochiai:
This PR fixes mismatch of `__attribute__((noreturn))` in `LogMessage::Fail()`'s header and source in Bazel build.

The problem seems to be caused by incomplete autoconf emulation in `glog.bzl`.

This remove the `"'noreturn' function does return"`  warning that can be a trouble to build with `'-Werror'`.


drigz:
Thank you so much for figuring this out @aochiai! This warning has been annoying me but I didn't know what was stopping the __attribute__ from working.

PR #313, opened 2018-04-25
Add BUILD_SHARED_LIBS option

DaikiMaekawa:


DaikiMaekawa:
I signed it.

sergiud:
Why is this needed?

DaikiMaekawa:
I wanted switch this value from my ccmake. Isn't it useful for others?

CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ set (CPACK_PACKAGE_VERSION ${PROJECT_VERSION})
 option (WITH_GFLAGS "Use gflags" ON)
 option (WITH_THREADS "Enable multithreading support" ON)
 option (WITH_TLS "Enable Thread Local Storage (TLS) support" ON)
+option (BUILD_SHARED_LIBS "Request build of shared libraries" OFF)
```
sergiud: I would change the description to a shorter "Build shared libraries" to be consistent with the existing descriptions.

DaikiMaekawa: addressed.

DaikiMaekawa:
@sergiud who do i talk to merge the PR?

sergiud:
@DaikiMaekawa I'll merge it.

andschwa:
@sergiud It's not really needed... but it does make the variable show up nicely. I think we have the same code in the Mesos build just for "aesthetics."

sergiud:
@andschwa I agree.

CMakeLists.txt
aesophor: @sergiud 
Currently, if glog is built with CMake, then the shared library is not built and installed by default, while this was the default behavior with autoconf and libtool (which were dropped recently). Could we set `BUILD_SHARED_LIBS` to ON by default?

Maybe some users are still using the shared library, and when they move to the recent version of glog, they'll probably be confused because the new version of .so is not installed by default.

Issue #279, opened 2017-12-24
[Bazel] config.h: No such file or directory


calder:
When building a Bazel project that imports glog, I get the following error:
```bash
ERROR: .../external/com_google_glog/BUILD:5:1: C++ compilation of rule '@com_google_glog//:glog' failed (Exit 1)
In file included from external/com_google_glog/src/utilities.h:75:0,
                 from external/com_google_glog/src/demangle.cc:38:
external/com_google_glog/src/base/mutex.h:105:64: fatal error: config.h: No such file or directory
compilation terminated.
```

This could be a regression in e6e2e1372aeb3f6ac6ca5f4a3d72bad8dce3c724 `bazel/glog.bzl`:
```diff
-            '/'.join([PACKAGE_NAME, 'config.h']) if PACKAGE_NAME else 'config.h',
+            'glog_internal/src/config.h',
```

Replacing `glog_internal/src/config.h` with `config.h` fixes the issue.

drigz:
This should be fixed by #277, could you try using the latest glog?

drigz:
Closing - please reopen if you're still seeing this with the latest glog.

jondo2010:
It looks like this is still a problem when building without gflags:

```
# Override the upstream BUILD file to disable gflags dependency
new_git_repository(
    name = "com_google_glog",
    remote = "https://github.com/google/glog.git",
    commit = "e0341da07bf7cdc00eafdc00744a807a7e5ba427", # Current head
    build_file_content = """
load(':bazel/glog.bzl', 'glog_library')
glog_library(with_gflags=False)
"""
)
```

```
INFO: Found 1 target...
ERROR: /home/john/.cache/bazel/_bazel_john/b74fab4f1d4785d9b546125b63063446/external/com_google_glog/BUILD.bazel:3:1: C++ compilation of rule '@com_google_glog//:glog' failed (Exit 1)
In file included from external/com_google_glog/src/utilities.h:75:0,
                 from external/com_google_glog/src/raw_logging.cc:34:
external/com_google_glog/src/base/mutex.h:105:64: fatal error: config.h: No such file or directory
compilation terminated.
```

PR #314, opened 2018-04-26
Fix Bazel build with_gflags=False

drigz:
as reported by @jondo2010:
https://github.com/google/glog/issues/279#issuecomment-384289681

This was caused by an operator precendence issue: with_gflags=False
would result in copts=[] and the build would not work. This has
presumably never been tested.

Unfortunately, it is not easy to test this in CI with the current
structure. The glog_library() macro can only be used once, as it
creates various rules with fixed names (eg :gen_sh), and must be used at
@com_google_glog//. The only way to use with_gflags=False is to
overwrite the root BUILD file.

### DIFF
--- a/bazel/glog.bzl
+++ b/bazel/glog.bzl
@@ -13,6 +13,37 @@ def glog_library(namespace='google', with_gflags=1):
     else:
         gendir = '$(GENDIR)'
 
+    glog_copts = [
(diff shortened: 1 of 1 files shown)


drigz:
@iirina for Bazel review.

drigz:
Looks like I forgot about this :/ but it was resolved by #339.

PR #322, opened 2018-05-28
feat: log rotate when file larger than FLAG_file_max_num

bartsimpson89:


src/logging.cc
```diff
@@ -961,6 +969,78 @@ bool LogFileObject::CreateLogfile(const string& time_pid_string) {
...
 }
 
+void LogFileObject::DeleteIfMaxNum(string &strDir)
+{
+	if (strDir.empty()) return;
+	string strSeverityName(GetLogSeverityName(severity_));
+	int fileMaxSizeNum = 0;
+#if defined(OS_WINDOWS)
+	long   hFile   =   0;
+	struct _finddata_t *earliestFile = (struct _finddata_t *) malloc(sizeof(struct _finddata_t));
```
sergiud: Is there a reason for heap allocation of `_finddata_t`?

bartsimpson89: sorry, i  misunderstand the struct of _finddata_t, for it may too large while out of the function stack

bartsimpson89: ok i will change this, thankyou

PR #327, opened 2018-06-18
add custom target 'check' for more verbose tests

NeroBurner:
adds a custom cmake target `check`. In case of an error in the testsuite the errors are printed to the terminal (instead of just telling the user which test failed). Very handy for CI systems

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -619,6 +619,10 @@ if (BUILD_TESTING)
   if (TARGET symbolize_unittest)
     add_test (NAME symbolize COMMAND symbolize_unittest)
   endif (TARGET symbolize_unittest)
+
+  add_custom_target(check ${CMAKE_COMMAND} -E env CTEST_OUTPUT_ON_FAILURE=1
+                    ${CMAKE_CTEST_COMMAND} -C $<CONFIG> --verbose
+                    WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 endif (BUILD_TESTING)
 
 install (TARGETS glog


sergiud:
Why is this needed? Can't you just call `ctest --output-on-failure` directly?

NeroBurner:
It's nice to work with targets in CI environments (when testing on Windows/Linux).
instead of creating the build directory and then change into the newly created folder to build and test one can use just `cmake` commands. Then `cmake` selects the right tool to compile (`make` on linux, `msvc` on windows)

For example in Linux one can create the build directory and call test by hand
```bash
mkdir _build
cd _build
cmake ..
make
ctest
```

or let cmake do the heavy lifting (on every platform):
```bash
cmake -H. -B_build
cmake --build _build
cmake --build _build --target check
```

No need to remember how to compile in the command line on Windows with `msvc` or `ninja`

Another nice effect of this commit is, that no matter if you compile on Visual Studio or Makefiles you can always use the target `check` in the commandline. Otherwise one would have to know that under Visual Studio the test target is called `RUN_TESTS` and on Makefiles the test target is called `test`

sergiud:
I understand all this. However, this is not a "one size fits all" solution since one user may prefer a complete output of the log, the other none at all, and the third just on failures. Hence, I'm questioning the usefulness of this PR.

NeroBurner:
I could add three custom targets `check_silent`, `check_failure`, `check_verbose` covering your examples. Would this be enough or would it add to much clutter?

sergiud:
The scenarios I mentioned are just examples. There are many other parameters that `ctest` provides to support different scenarios.

I believe adding multiple check targets that cover the aforementioned cases would add too much noise without any real benefit. Especially, since there's already a standard test/RUN_TESTS target, albeit the verbose output.

Finally, CI runners typically require providing the command line for building and testing projects anyhow. Hence, calling `ctest` directly is generally not an issue.

NeroBurner:
Thanks for the explanation.

I found another workaround using cmake targets
```
CTEST_OUTPUT_ON_FAILURE=1 cmake --build _build --target test
```

PR #339, opened 2018-07-04
copts are passed correctly when building w/o gflags

### LINKED ISSUE #338: Build fails when compiling w/o GFLAGS
Build fails when setting `with_gflags=0` in file `BUILD`

```
ERROR: /root/.cache/bazel/_bazel_root/02effb97cc26fc72f32cf2bfdfcd0b54/external/glog/BUILD:5:1: C++ compilation of rule '@glog//:glog' failed (Exit 1)
In file included from external/glog/src/utilities.h:75:0,
                 from external/glog/src/vlog_is_on.cc:35:
external/glog/src/base/mutex.h:105:64: fatal error: config.h: No such file or directory
```

This is caused because of a bug in `glog.bzl` (`copts` are not set correctly when not using gflags)

I'll open a PR to fix this

levosos:
resolves #338

### DIFF
--- a/bazel/glog.bzl
+++ b/bazel/glog.bzl
@@ -73,10 +73,10 @@ def glog_library(namespace='google', with_gflags=1):
 
             # Include generated header files.
             '-I%s/glog_internal' % gendir,
-        ] + [
+        ] + ([
             # Use gflags to parse CLI arguments.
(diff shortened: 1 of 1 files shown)


levosos:
I signed it!

drigz:
@levosos Thanks for fixing this!

PR #347, opened 2018-07-18
Fix stacktrace on Bazel build.

### LINKED ISSUE #346: Stacktrace is not printed by failure signal handler when built with Bazel
When glog is built with Bazel, failure signal handler is unable to print stack trace.

*** Aborted at 1531880485 (unix time) try "date -d @1531880485" if you are using GNU date ***
PC: @                0x0 (unknown)
zsh: segmentation fault  bazel run :hello

On Linux, no stacktrace is shown. On macOS, it enters deadlock for some reason and can be killed only with SIGKILL (no Ctrl+C), which is worse.

nya3jp:
We need at least following defs to be set to print stacktrace in
failure signal handler.

- HAVE_UNWIND_H: On Linux and macOS. unwind.h is usually present
  by default on those systems.
- HAVE_DLADDR: On macOS.

Note that today glog does not build with Bazel on Windows.

Resolves #346.

### DIFF
--- a/bazel/glog.bzl
+++ b/bazel/glog.bzl
@@ -73,7 +73,16 @@ def glog_library(namespace='google', with_gflags=1, **kwargs):
 
             # Include generated header files.
             '-I%s/glog_internal' % gendir,
-        ] + ([
+        ] + select({
+            # For stacktrace.
+            '@bazel_tools//src/conditions:darwin': [
+                '-DHAVE_UNWIND_H',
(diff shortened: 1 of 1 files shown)


glalonde:
This would be super cool.

ukai:
sorry for late.

seems PR is too old for buildkite verification.
could you rebase to master to pick up the changes to the WORKSPACE file and others?

nya3jp:
Sure, rebased.

PR #364, opened 2018-09-05
Added support for latest gflags.

sumit-g:
config.h conflict seems to be preventing us from moving to latest gflags release. This patch tries to fix that by renaming glog's config.h to glog_config.h

Also, gflags dependency was hardcoded to be @com_github_gflags_gflags. This patch makes it a variable which can be changed by the root level BUILD file.


CMakeLists.txt
```diff
@@ -410,7 +410,7 @@ endif (WITH_THREADS AND Threads_FOUND)
 
 set (TEST_SRC_DIR \"${CMAKE_CURRENT_SOURCE_DIR}\")
 
-configure_file (src/config.h.cmake.in config.h)
```
sergiud: Please put the config under `glog/config.h` and reference it as such.

drigz:
@sumit-g Thanks for the PR! I'm reviewing this from the Bazel side, I don't know about the CMake build.

Could you explain in more detail why you're making this change to version in WORKSPACE? https://github.com/gflags/gflags/commit/77592648e3f3be87d6c7123eb81cbad75f9aef5a is more recent that the v2.2.1 release (which is from July 2017), and includes various fixes - most importantly https://github.com/gflags/gflags/commit/57ceb0ecc8bf5c129bf99fa9472b25342daecf9e, which avoids leaking the private config.h header to all libraries that depend on glog or gflags. Perhaps a better solution for you would be to upgrade @com_github_gflags_gflags to a more recent commit? Unfortunately there is no more recent release than v2.2.1, but 775926 is certainly an improvement upon v2.2.1 when using Bazel.

Regarding the `gflags_dep` parameter, this seems unrelated to the header version bump - could you split it into a separate pull request? Also, please document its behavior in the comment at the top of the file - in particular, that the user needs to provide `with_gflags=1` in addition to `gflags_dep`.

sumit-g:
@drigz Thanks for the review. I am basically making glog immune to any config.h leakage issues (current or future). In terms of which gflags commit should glog depend upon, I think its best to stay with a released version.

I have removed the gflags_dep change. I will use a different PR for that.

CMakeLists.txt
sumit-g: There is already a src/glog/ directory. Creating another glog/ directory under root will make it more confusing as to why some headers are under src/glog/ and some are under glog/ . I think glog_config.h is more readable IMHO.

drigz:
> I am basically making glog immune to any config.h leakage issues (current or future).

This seems unnecessary. glog only has one dependency, gflags, which is fairly stable (11 commits in the last year). I don't see a risk of config.h leakage in the future.

> In terms of which gflags commit should glog depend upon, I think its best to stay with a released version.

I agree that that would be ideal - I've filed https://github.com/gflags/gflags/issues/256 to ask for a release that doesn't leak config.h. However, given that 2.2.1 is not good for Bazel users (even if glog were immune to config.h leaking, other libraries depending on glog might not be) I think it's better to depend on the more recent code for now.

CMakeLists.txt
sergiud: The `glog/config.h` should be put into the build directory.

PR #365, opened 2018-09-10
Allow adding vmodules at runtime - update cached site levels accordingly 

romange:


src/logging.cc
```diff
@@ -1153,7 +1153,9 @@ static LogMessage::LogMessageData fatal_msg_data_shared;
 // LogMessageData object exists (in this case glog makes zero heap memory
 // allocations).
 static GLOG_THREAD_LOCAL_STORAGE bool thread_data_available = true;
-static GLOG_THREAD_LOCAL_STORAGE char thread_msg_data[sizeof(LogMessage::LogMessageData)];
+static GLOG_THREAD_LOCAL_STORAGE std::aligned_storage<
+    sizeof(LogMessage::LogMessageData),
+    alignof(LogMessage::LogMessageData)>::type thread_msg_data;
```
sergiud: It seems you mixed several (unrelated) changes in this PR (see #374).

romange: Sorry about this. Fixed

romange:
PTAL

romange:
Hi @sergiud I remove the unrelated change. Please note that I've sent another PR to solve the same problem #366 
The difference between them is that in this PR the main performance impact is concentrated inside SetVLOGLevel function - I traverse all VLOG cached instance to update their value.

#366 I think it's less intrusive but the performance impact is distributed upon all VLOG calls *after* SetVLOGLevel has been called since I invalidate caching for all VLOG statements.

I think #366 proposes less intrusive solution if SetVLOGLevel is not called at all because the size of VLOG state is smaller.

PR #366, opened 2018-09-12
Propose another way of solving the InitVModule bug 

romange:
where new vmodules do not invalidate cached vlog levels


romange:
@sergiud  Can you please take a look?

romange:
@sergiud  This PR handles a bug where SetVLOGLevel is called after module verbosity level has been cached. In that case the level is not updated. I've changed this behavior by triggering a refresh using a  sequence (version) number.

PR #370, opened 2018-10-12
Fix warnings

hemantbits:
Fixed these warnings:

```
src/logging_unittest.cc: At global scope:
src/logging_unittest.cc:1081:13: warning: 'void MyCheck(bool, bool)' defined but not used [-Wunused-function]
 static void MyCheck(bool a, bool b) {
             ^~~~~~~
src/logging_unittest.cc:1078:13: warning: 'void MyFatal()' defined but not used [-Wunused-function]
 static void MyFatal() {
             ^~~~~~~
```

### DIFF
--- a/src/logging_unittest.cc
+++ b/src/logging_unittest.cc
@@ -1075,13 +1075,14 @@ TEST(Strerror, logging) {
 
 // Simple routines to look at the sizes of generated code for LOG(FATAL) and
 // CHECK(..) via objdump
+/*
 static void MyFatal() {
   LOG(FATAL) << "Failed";
 }
 static void MyCheck(bool a, bool b) {
   CHECK_EQ(a, b);
 }
-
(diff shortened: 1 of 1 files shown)


src/logging_unittest.cc
```diff
@@ -1075,13 +1075,14 @@ TEST(Strerror, logging) {
...
 // CHECK(..) via objdump
+/*
 static void MyFatal() {
   LOG(FATAL) << "Failed";
 }
 static void MyCheck(bool a, bool b) {
   CHECK_EQ(a, b);
 }
-
+*/
```
sergiud: What's the reason for keeping the functions in a comment?

hemantbits: I was thinking if someone might want to use them in future. If you think they should be removed, please let me know

Issue #212, opened 2017-07-17
Data race in raw_logging.cc?


adityamarella:
Helgrind shows a possible datarace with the following stack trace

```
==18== Possible data race during write of size 8 at 0xD01620 by thread #1
==18== Locks held: none
==18==    at 0x57AFC9: memcpy (string3.h:51)
==18==    by 0x57AFC9: google::RawLog__SetLastTime(tm const&, int) (raw_logging.cc:78)
==18==    by 0x572DC9: google::LogMessage::Init(char const*, int, int, void (google::LogMessage::*)()) (logging.cc:1217)
==18==    by 0x573491: google::LogMessage::LogMessage(char const*, int, int, int, void (google::LogMessage::*)()) (logging.cc:1142)
```
After going through the code, this indeed seems like a data race. What happens if multiple threads call `google::RawLog__SetLastTime` at the same time?

adityamarella:
@shinh could you please comment on this. Thanks!

shinh:
Indeed this is a race. It seems Google's internal version has removed datetime from raw log at all. If no one objects, I'd remove it from glog, too.

Same report: https://code.google.com/archive/p/google-glog/issues/62

adityamarella:
Thanks for confirming that this is an issue. I have no objection to removing datetime from raw log, in fact I am patching my local copy to fix this.

shinh:
I was reading the internal code wrongly. It seems the internal code still outputs timestamp at least on Linux. Hmm.

PR #377, opened 2018-10-29
remove time from raw logging

plmuon:
Fix for issue #212.

Thread sanitizer, mandatory in our ci process, refuses to accept glog because of this issue...

As the discussion in issue #212 indicated that google has already removed time from raw logging because of the same issue, I did the same. Instead, we log zeros (to keep the format the same as the normal logger).


plmuon:
I signed the CLA.

PR #388, opened 2018-11-13
Make symbolize.cc thread safe even on shared fds

tzik:
ReadFromOffset in symbolize.cc used to call lseek() + read() to read
data from fd. However, the fd may be reused for multiple symbolize
requests from multiple threads, and causes a race around the fd read
offset.

This updates it to use pread() to resolve the race.

### DIFF
--- a/src/symbolize.cc
+++ b/src/symbolize.cc
@@ -136,17 +136,20 @@ _END_GOOGLE_NAMESPACE_
 
 _START_GOOGLE_NAMESPACE_
 
-// Read up to "count" bytes from file descriptor "fd" into the buffer
-// starting at "buf" while handling short reads and EINTR.  On
-// success, return the number of bytes read.  Otherwise, return -1.
-static ssize_t ReadPersistent(const int fd, void *buf, const size_t count) {
+// Read up to "count" bytes from "offset" in the file pointed by file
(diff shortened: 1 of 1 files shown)


tzik:
JFYI @hajimehoshi

src/symbolize.cc
```diff
@@ -136,17 +136,20 @@ _END_GOOGLE_NAMESPACE_
...
+static ssize_t ReadFromOffset(const int fd, void *buf, const size_t count,
+                              const off_t offset) {
   SAFE_ASSERT(fd >= 0);
   SAFE_ASSERT(count <= std::numeric_limits<ssize_t>::max());
   char *buf0 = reinterpret_cast<char *>(buf);
   ssize_t num_bytes = 0;
   while (num_bytes < count) {
     ssize_t len;
-    NO_INTR(len = read(fd, buf0 + num_bytes, count - num_bytes));
+    NO_INTR(len = pread(fd, buf0 + num_bytes, count - num_bytes,
```
ukai: is it available on windows?

tzik: Not sure, but IIUC, this is not used on Windows as we have `defined(OS_WINDOWS) || defined(OS_CYGWIN)` branch below.

ukai: ah, ok.

PR #391, opened 2018-11-14
Upstream Chromium local changes to symbolize.cc

tzik:
Chromium has its own fork of symbolize.cc and symbolize.h, and it has
several not-yet-upstreamed changes, that are not specific to Chromium.

This patch upstreams such a changes.

  Fixed google::FindSymbol reading past end of a section
  https://chromium.googlesource.com/chromium/src/+/3dae0a2d7da309159a301e40a592692ec8431a38

  Fix -INT_MIN integer overflow in itoa_r().
  https://chromium.googlesource.com/chromium/src/+/ac4d28e9cb934e01ff16a9e67a5f387caa82719f

  Add print_unsymbolized_stack_traces gn arg.
  https://chromium.googlesource.com/chromium/src/+/6a2726776f59cdcf0090f584a15db4a1ab8c40d7

  Switch to standard integer types in base/.
  https://chromium.googlesource.com/chromium/src/+/9b6f42934e5a1e65ebfc668d91a28a6e2678a14c

### DIFF
--- a/src/symbolize.cc
+++ b/src/symbolize.cc
@@ -58,6 +58,7 @@
 
 #include <string.h>
 
+#include <algorithm>
 #include <limits>
 
 #include "symbolize.h"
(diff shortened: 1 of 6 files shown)


src/symbolize.cc
```diff
@@ -780,8 +781,13 @@ static ATTRIBUTE_NOINLINE bool SymbolizeAndDemangle(void *pc, char *out,
                                                              out_size - 1);
   }
 
+#if defined(PRINT_UNSYMBOLIZED_STACK_TRACES)
```
ukai: could we add this for autoconf and/or cmake?

tzik: Sure. Updated.

PR #412, opened 2018-12-26
Fix unused variable warning in GCC

pwnall:
`kAlign` is only used the `#ifdef` / `#else` branch, so GCC complains if both `GLOG_THREAD_LOCAL_STORAGE` and `HAVE_ALIGNED_STORAGE` are defined.


sergiud:
@pwnall You're welcome!

PR #413, opened 2018-12-26
Fix bazel build with gflags

prem-nuro:
The code references `@ac_cv_have_libgflags@` but not `@ac_cv_have_libgflags_h@`. This corrects that.

`int(with_gflags)` incorporates the possibility of `True/False`: https://github.com/bazelbuild/bazel/issues/4792


prem-nuro:
I signed it!

prem-nuro:
Sorry, I have some email aliases and we just set up corporate CLA, so not sure what the fix is here.

Issue #14, opened 2015-04-08
gcc 4.7.2, 4.8.3, 4.9.2 unit test failures on C++11


pshahid:
Hi, I've encountered some issues with the demangle unit tests using glog-0.3.4 on the gcc family of compilers. This is being run on fedora 20, with the compilers as seen in the title, and with C++11. 

What follows is the last 30ish lines from the make/make check output.
### gcc 4.7.2

```
./src/logging_striplog_test.sh
In DBG mode; not checking strings
PASS
./demangle_unittest  # force to create lt-demangle_unittest
./src/demangle_unittest.sh
Mangled symbols (14 out of 801) found in demangle.dm:
_ZN9__gnu_cxx13new_allocatorIPN6google7LogSinkEE9constructIS3_JRKS3_EEEvPT_DpOT0_
_ZN9__gnu_cxx13new_allocatorISsE9constructISsJRKSsEEEvPT_DpOT0_
_ZN9__gnu_cxx13new_allocatorISsE9constructISsJSsEEEvPT_DpOT0_
_ZNSt16allocator_traitsISaIPN6google7LogSinkEEE12_S_constructIS2_JRKS2_EEENSt9enable_ifIXsrNS4_18__construct_helperIT_JDpT0_EEE5valueEvE4typeERS3_PSA_DpOSB_
_ZNSt16allocator_traitsISaIPN6google7LogSinkEEE9constructIS2_JRKS2_EEEvRS3_PT_DpOT0_
_ZNSt16allocator_traitsISaISsEE12_S_constructISsJRKSsEEENSt9enable_ifIXsrNS1_18__construct_helperIT_JDpT0_EEE5valueEvE4typeERS0_PS7_DpOS8_
_ZNSt16allocator_traitsISaISsEE12_S_constructISsJSsEEENSt9enable_ifIXsrNS1_18__construct_helperIT_JDpT0_EEE5valueEvE4typeERS0_PS5_DpOS6_
_ZNSt16allocator_traitsISaISsEE9constructISsJRKSsEEEvRS0_PT_DpOT0_
_ZNSt16allocator_traitsISaISsEE9constructISsJSsEEEvRS0_PT_DpOT0_
_ZNSt6vectorIPN6google7LogSinkESaIS2_EE19_M_emplace_back_auxIJRKS2_EEEvDpOT_
_ZNSt6vectorISsSaISsEE12emplace_backIJSsEEEvDpOT_
[...]

PR #50, opened 2015-08-13
Fix ABI demangling for the GCC 5.x case.

Jo-Con-El:
When glog is compiled with gcc-5.2 in cxx11 ABI mode, it barfs about unmangled symbols.  This patches it getting inspiration from binutils and demangle.cc itself, although it may be totally wrong or maybe have to use ParseAbiTag in more places.  I haven't read the spec for the symbols, though.

Symbols are demangled now as

```
google::LogDestination::addresses_[cxx11]
```

This patch makes the demangle unit test pass correctly (as there's no more symbols starting with _Z), so closes #40.

PR #414, opened 2019-01-02
Fix demangling template parameter packs

tanderson-google:
Clang 4.0.1-10 and gcc 7.3.0 both mangle the function "void add<int>(int)" as
"_Z3addIJiEEvDpT_".  The template parameter pack is of the form
J <template-arg>* E

The opening character for a param pack could be either I or J, as libiberty
follows [1].  This change simply adds the J case.

[1] https://github.com/gcc-mirror/gcc/blob/fbd263526ad105a953fd51d9f7bca2c3f268cf82/libiberty/cp-demangle.c#L3209

### DIFF
--- a/src/demangle.cc
+++ b/src/demangle.cc
@@ -1097,10 +1097,11 @@ static bool ParseTemplateArgs(State *state) {
 // <template-arg>  ::= <type>
 //                 ::= <expr-primary>
 //                 ::= I <template-arg>* E        # argument pack
+//                 ::= J <template-arg>* E        # argument pack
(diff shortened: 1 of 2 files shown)


hlsyounes:
I believe this should fix #14 too. The mangled strings all have "J" in them in that issue. This is different from #50, which would have "B" in the mangled strings.

tanderson-google:
> How about adding the examples to demangle_unittest.txt instead? Then they would be tested by the FromFile test case.

Done

PR #425, opened 2019-01-29
remove time from raw logging

plmuon:
Fix for issue #212.

The issue caused trouble in my application, so I had to get rid of time in raw logging, like suggested by https://github.com/google/glog/issues/212#issuecomment-315989830


src/raw_logging.cc
```diff
@@ -120,16 +109,13 @@ void RawLog__(LogSeverity severity, const char* file, int line,
     return;  // this stderr log message is suppressed
   }
   // can't call localtime_r here: it can allocate
```
ukai: can remove this comment?

src/raw_logging.cc
```diff
@@ -120,16 +109,13 @@ void RawLog__(LogSeverity severity, const char* file, int line,
...
-  struct ::tm& t = last_tm_time_for_raw_log;
   char buffer[kLogBufSize];
   char* buf = buffer;
   int size = sizeof(buffer);
 
   // NOTE: this format should match the specification in base/logging.h
```
ukai: do we still need to match the format?

abseil's internal raw_logging doesn't include timestamp.
https://github.com/abseil/abseil-cpp/blob/master/absl/base/internal/raw_logging.cc#L153

src/raw_logging.cc
plmuon: Yes of course, it overlooked it.

src/raw_logging.cc
plmuon: I just followed the NOTE comment, but I would also think that it is not necessary to keep the format the same.

Btw. I don't know what happened with appveyor VS builds (mingw was ok). I'd have some trouble to get a windows/VS environment and try to reproduce.

src/raw_logging.cc
```diff
@@ -120,16 +109,13 @@ void RawLog__(LogSeverity severity, const char* file, int line,
...
-  DoRawLog(&buf, &size, "%c%02d%02d %02d:%02d:%02d.%06d %5u %s:%d] RAW: ",
+  DoRawLog(&buf, &size, "%c0000 00:00:00.000000 %5u %s:%d] RAW: ",
```
shinh: Ukai-san, could you check if the internal version has these numbers? If the internal version has removed them, we should do the same, too.

shinh:
Sorry for the latency.

src/raw_logging.cc
ukai: it uses abseil-cpp's raw_logging

https://github.com/abseil/abseil-cpp/blob/master/absl/base/internal/raw_logging.cc

Issue #335, opened 2018-07-02
Unittest generates warning with C++17


NeroBurner:
When compiling with C++17 enabled gcc throws the following warning:
```
[ 76%] Building CXX object CMakeFiles/utilities_unittest.dir/src/utilities_unittest.cc.obj
/usr/src/app/src/logging_unittest.cc: In function 'void TestDCHECK()':
/usr/src/app/src/logging_unittest.cc:575:3: warning: 'template<class> class std::auto_ptr' is deprecated [-Wdeprecated-declarations]
   auto_ptr<int64> sptr(new int64);
   ^~~~~~~~
In file included from /usr/lib/gcc/x86_64-w64-mingw32/7.3-win32/include/c++/bits/locale_conv.h:41:0,
                 from /usr/lib/gcc/x86_64-w64-mingw32/7.3-win32/include/c++/locale:43,
                 from /usr/lib/gcc/x86_64-w64-mingw32/7.3-win32/include/c++/iomanip:43,
                 from /usr/src/app/src/logging_unittest.cc:44:
/usr/lib/gcc/x86_64-w64-mingw32/7.3-win32/include/c++/bits/unique_ptr.h:51:28: note: declared here
   template<typename> class auto_ptr;
                            ^~~~~~~~
```

For a full build output see the following log: https://travis-ci.org/NeroBurner/glog/jobs/399043400

With MSVC 2017 and C++17 enabled the build throws an error (full log https://ci.appveyor.com/project/NeroBurner/glog/build/1.0.111/job/dl21jws89k12c66u#L266)
```
ClCompile:
[...]

PR #428, opened 2019-02-01
Stop using auto_ptr

shinh:
Saw an error in AppVeyor:

c:\projects\glog\src\logging_unittest.cc(575): error C2065: 'auto_ptr': undeclared identifier [C:\projects\glog\_build_vs-15-2017-win64-cxx17_Debug\logging_unittest.vcxproj]

### DIFF
--- a/src/logging_unittest.cc
+++ b/src/logging_unittest.cc
@@ -572,9 +572,10 @@ void TestDCHECK() {
   DCHECK_GT(2, 1);
   DCHECK_LT(1, 2);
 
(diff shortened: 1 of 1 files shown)


shinh:
and this is the issue: https://github.com/google/glog/issues/335

shinh:
Travis failed for std::unique_ptr. We might want to use -std=c++11 or something, but just using raw pointers should be also OK. It's just about a unittest anyway. auto_ptr related issue has been gone now.

PR #431, opened 2019-02-16
 Protect macros from user code to the left of them

romkatv:
Currently the following code compiles (and the check passes):

```c++
!CHECK(2 + 2 == 5);
```

Such mistakes can happen by accident. They can be especially difficult to notice if the user code that interferes with the CHECK is on the previous line:

```c++
DoStuff(42);~

// Make sure the index is within bounds. Otherwise a malicious
// actor could execute arbitrary code on our machine.
CHECK(index < length);
```

Oops! The check always passes.

This PR causes this code to be rejected at compile time.


shinh:
Interesting trick! LGTM

Issue #423, opened 2019-01-27
Glog should delete old log files automaticly


DinoStray:
For example, I wanna keep log files for only 7 days.
Then the  glog should delete all the log files which last modify time before 7 days.

aesophor:
I think a cron job can do it

This deletes the logs produced by glog under /tmp that are last modified 7 days ago.
```
$ find /tmp -name 'YourProjectName.*' -mtime +7 -delete
```

Then `$ crontab -e` and add this line to run it daily at 00:00 AM
```
0 0 * * * find /tmp -name 'YourProjectName.*' -mtime +7 -delete
```

DinoStray:
> I think a cron job can do it
> 
> This deletes the logs produced by glog under /tmp that are last modified 7 days ago.
> 
> ```
> $ find /tmp -name 'YourProjectName.*' -mtime +7 -delete
> ```
> Then `$ crontab -e` and add this line to run it daily at 00:00 AM
> 
> ```
> 0 0 * * * find /tmp -name 'YourProjectName.*' -mtime +7 -delete
> ```

But If I use docker or  k8s, there is no crontab.
Clean old logs should done by the glog itself

aesophor:
@DinoStray I've written the code to automatically remove old logs, but they are not yet integrated into the original codebase.
@sergiud I'm not sure when to call it. Calling it at the end of LogFileObject::FlushUnlocked() seems really unwise, since that scans for old logs too frequently. Could you please help me with this? Thank you!

letrthong:
Raspberry doesn't have real time clock so can not get timestamp. Please check the case

letrthong:
i wrote code to limit maximum size of all  files about  3M which is running on raspberry pi.

aesophor:
@DinoStray I tried to implement this and it works on my linux box! :smile: (gentoo-4.19.66) But I haven't tested it on Windows, I'll do it once I have spare time.

---

Simply add this to your project to enable **log cleaner**, it is disabled by default. Once enabled, it will automatically remove old logs that are last modified over 3 days. Only the logs produced by the project which glog is linked in will be removed, logs from other projects won't be touched.
```
google::EnableLogCleaner(3);
```

You can also disable it at anytime by calling
```
google::DisableLogCleaner();
```

Please see here for technical details.
https://github.com/google/glog/pull/432#issuecomment-523680414

aesophor:
[...]

PR #432, opened 2019-02-17
Add support for automatic removal of old logs

aesophor:
This PR is in regard to #423. The code has been added and tested , but not yet integrated into glog. 

Example usage:
```
vector<string> overdue_log_names = GetOverdueLogNames("/tmp", 1);

for (const auto& name : overdue_log_names) {
  unlink(name.c_str());
}
```

GetOverdueLogNames() will check all filenames under log_directory, and return a list of files whose last modified time is over the given days ( calculated with `difftime()` ), so that we can simply for unlink() the filenames in the returned vector.

However, I'm not sure when is the right time to call it.


aesophor:
I signed it!

Steven0917:
Guess the patch was implemented based on Linux. For Windows user, https://github.com/tronkko/dirent will be needed.

src/logging.cc
```diff
@@ -1222,9 +1227,11 @@ void LogFileObject::Write(bool force_flush,
     }
 #endif
     // Perform clean up for old logs
-    for (const auto& dir : GetLoggingDirectories()) {
-      for (const auto& name : GetOverdueLogNames(dir, 3)) {
-        static_cast<void>(unlink(name.c_str()));
+    if (log_cleaner_enabled_) {
+      for (const auto& dir : GetLoggingDirectories()) {
+        for (const auto& name : GetOverdueLogNames(dir, 3)) {
```
aesophor: This hardcoded day count has been fixed in a later commit.

src/logging.cc
```diff
@@ -1136,83 +1221,16 @@ void LogFileObject::Write(bool force_flush,
...
-  if ((dir=opendir(log_directory.c_str()))) {
-    while ((ent=readdir(dir))) {
-      string filename = log_directory + string(ent->d_name);
-      if (IsGlogLog(filename) && LastModifiedOver(filename, days)) {
-        overdue_log_names.push_back(filename);
+    // Perform clean up for old logs
+    for (const auto& dir : GetLoggingDirectories()) {
+      for (const auto& name : GetOverdueLogNames(dir, 3)) {
+        static_cast<void>(unlink(name.c_str()));
       }
```
aesophor: This part in diff might seems confusing. Basically:
1. All these functions (which are related to log cleaner) are moved upper
2. Line 1224-1227: This two-level for loop will be run whenever a flush is performed. https://github.com/google/glog/blob/7af21ae7af57d7bf5ca3c13e2ffa7fd5c37ed698/src/logging.cc#L1225

aesophor:
> Guess the patch was implemented based on Linux. For Windows user, https://github.com/tronkko/dirent will be needed.

Thank you very much for your help! CI has passed on windows!
This is what I've done currently:

To enable log cleaner, in your project (log cleaner is disabled by default).
```
google::EnableLogCleaner(3);
```
And then your project will check if there are overdue logs whenever a flush is performed. In this example, the logs whose last modified times are >= 3 will be unlink(). It calls GetLoggingDirectories() to look for overdue logs.

You may also disable it at anytime.
```
google::DisableLogCleaner();
```

I try to check if a file is **a glog file produced by the project it is linked in** carefully. It will only remove a log if its filename matches: `<program name>.<hostname>.<user name>.log.<severity level>`, so logs from other projects should be untouched.
https://github.com/google/glog/blob/b8a61d70834977d18a524fc849ada2f6dea0d189/src/logging.cc#L871

src/logging.cc
```diff
@@ -833,6 +833,91 @@ void LogDestination::DeleteLogDestinations() {
...
+    if (t.length() > 0) {
+      tokens.push_back(t);
+    }
+  }
+  return tokens;
+}
+
+bool IsGlogLog(const string& log_name) {
+  static const int kGlogFilenameTokenCount = 6;
+  vector<string> log_name_tokens = SplitString(log_name, '.');
```
aesophor: If `log_name` contains a `.`, then this will be broken

aesophor: Resolved by passing `ent->d_name` into IsGlogLog() directly so that we don't have to perform filename stripping ourselves, which also solves this problem.

src/logging.cc
```diff
@@ -827,6 +833,91 @@ void LogDestination::DeleteLogDestinations() {
...
```
aesophor: This is broken on windows, e.g., `windoze.exe`.
Splitting filename by `.` is not reliable.

aesophor: Resolved in 77a1f39 by matching the filename keyword by keyword, please see the commit message for details.

aesophor:
@sergiud Sorry for asking but any chance this patch could be merged?

I've tested them on linux (4.19.66-gentoo) and windows 10, old log files can be removed automatically. This feature is disabled by default.

cswuyg:
"It will only remove a log if its filename matches: <program name>.<hostname>.<user name>.log.<severity level>, so logs from other projects should be untouched."
>>> @aesophor  If I use SetLogDestination API to change base_filename_, old logs will not delete automatically.  is a bug?

aesophor:
> > "It will only remove a log if its filename matches: ...log., so logs from other projects should be untouched."
> 
> @aesophor If I use SetLogDestination API to change base_filename_, old logs will not delete automatically. is a bug?

@cswuyg Yes, this is a bug. I'm working on a patch for this.

aesophor:
@cswuyg Fixed in #502

PR #433, opened 2019-02-20
Use platform APIs to get username instead of env variables.

PawanKartikS:
Previously, username was obtained by reading the value of env variables. This PR implements the usage of platform based APIs to get the username.

### DIFF
--- a/src/utilities.cc
+++ b/src/utilities.cc
@@ -56,6 +56,10 @@
 
 #include "base/googleinit.h"
 
+#if defined(OS_WINDOWS)
+#include <Windows.h>    // GetUserNameA
+#endif
+
 using std::string;
 
 _START_GOOGLE_NAMESPACE_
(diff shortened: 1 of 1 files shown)


PawanKartikS:
I signed it!

PawanKartikS:
Travil check fail: Travis seems to be including `Windows.h` on Linux platforms when building with `CMake wine stable`.

src/utilities.cc
```diff
@@ -296,13 +300,17 @@ const string& MyUserName() {
   return g_my_user_name;
 }
 static void MyUserNameInitializer() {
-  // TODO(hamaji): Probably this is not portable.
+  int err = 0;
+  char user[1024];
 #if defined(OS_WINDOWS)
-  const char* user = getenv("USERNAME");
+  DWORD bufsize = sizeof(user);
```
sergiud: This does not look right.

sergiud: `user` could be `NULL`.

PawanKartikS: I apologize for the delay.
Please look at the complete diff.

`user` was originally of type `const char*`.  I made changes so it is now a string (as required by the APIs) - `char user[1024];`.

So `sizeof(user)` now actually returns `sizeof(char) * 1024`.

PR #437, opened 2019-02-27
Add extra information in log file header (application build/version, actual duration time)

d-uspenskiy:
It is very useful to have application version/build specific information in log file. 
It is possible to write in on start, but in case of log file rolling this information will be lost. 
Also it is useful to know actual running time of application when customer sends you just last log file


d-uspenskiy:
I signed it!

Issue #307, opened 2018-04-11
LogSink::ToString has no microsecond precision.


bmahler:
I noticed a custom LogSink was logging without microsecond precision:

```
I0330 16:05:21.000000 21188 master.cpp:8832] ...
I0330 16:05:21.000000 21188 master.cpp:8832] ...
I0330 16:05:21.000000 21188 master.cpp:8832] ...
```

Currently `LogSink::ToString` always uses 0 for the microsecond component of the timestamp. There is a FIXME related to this:

```
  // FIXME(jrvb): Updating this to use the correct value for usecs
  // requires changing the signature for both this method and
  // LogSink::send().  This change needs to be done in a separate CL
  // so subclasses of LogSink can be updated at the same time.
  int usecs = 0;
```
https://github.com/google/glog/blob/v0.3.5/src/logging.cc#L1631-L1635

Per the FIXME, it looks like the only way to fix this is by updating the signature of `LogSink::send`. Any third party LogSink implementations would need to have their signature adjusted. Would such a PR be acceptable?

bmahler:
@sergiud I'm happy to send a PR for this, would you be willing to help me get it merged?

sergiud:
@bmahler Sure, go ahead!

sergiud:
Fixed in 5792d60d68a8a7f60f06bc2b56966a94e18bdeb8.

PR #441, opened 2019-03-26
Extend the LogSink interface to be able to pass microseconds

asekretenko:
The goal of this PR is to solve the problem of passing microseconds into custom LogSink implementations that was described in https://github.com/google/glog/issues/307.

I am proposing a slightly different approach, however: instead of modifying the existing send() signature I am adding  the second one. 
This has an advantage of not breaking all the third-party LogSink implementations immediately (including ours) ... and, unfortunately, all the disadvantages associated with this approach.

### DIFF
--- a/src/glog/logging.h.in
+++ b/src/glog/logging.h.in
@@ -1426,6 +1426,16 @@ class GOOGLE_GLOG_DLL_DECL LogSink {
   virtual void send(LogSeverity severity, const char* full_filename,
                     const char* base_filename, int line,
                     const struct ::tm* tm_time,
+                    const char* message, size_t message_len, int32 usecs) {
+    send(severity, full_filename, base_filename, line,
+         tm_time, message, message_len);
+  }
+  // This send() signature is obsolete.
+  // New implementations should define this in terms of
+  // the above send() method.
+  virtual void send(LogSeverity severity, const char* full_filename,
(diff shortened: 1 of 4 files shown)


asekretenko:
I signed it!

jpeach:
@asekretenko would it make sense for the `usecs` parameter to be unsigned?

asekretenko:
> @asekretenko would it make sense for the `usecs` parameter to be unsigned?

Probably it makes sense. 
I cannot remember any way to obtain a negative value for `(x - static_cast<some_interger_type>(x)) * 1e-6` for an IEEE754-compliant double `x` if `x>0`. And you?

asekretenko:
> @asekretenko would it make sense for the `usecs` parameter to be unsigned?

On the other hand, on some crazy (or "slightly buggy") platform we might get a negative value here for some corner case.  Seeing ".-000001" in the microseconds would help to understand what the hell is going on much better than seeing  seemingly normal values in a strange order. Been there, done that (albeit with another logging system).

Probably it would be better to follow advice from [Google's style guide ](https://google.github.io/styleguide/cppguide.html) 
> Do not use an unsigned type merely to assert that a variable is non-negative.

and leave microseconds signed.

asekretenko:
@shinh @sergiud May I ask you to have a look at this PR?

PR #447, opened 2019-04-29
Fix indeterministic logging tests

### LINKED ISSUE #201: Test stl_logging fails in macOS CMake build
Running into a test failure on macOS 10.10 using AppleClang 6.1.0.6020053 with CMake when building glog 0.3.5. [1] When building with regular Make using the same macOS configuration, the same failure does not occur. [2] Neither CMake nor Make fail on Linux. Perhaps there is a bug in the CMake build w.r.t. macOS?

1: https://travis-ci.org/conda-forge/glog-feedstock/builds/246915696#L466
2: https://travis-ci.org/conda-forge/glog-feedstock/builds/246912247#L540

xhochy:
Printing the content of hash-based containers doesn't lead always to the same string as their ordering is not fixed. Luckily the number of possible permutations is so small that we can enumerate all of them.

Fixes #201.

### DIFF
--- a/src/stl_logging_unittest.cc
+++ b/src/stl_logging_unittest.cc
@@ -33,6 +33,7 @@
 
 #include <functional>
 #include <iostream>
+#include <iterator>
 #include <map>
 #include <ostream>
 #include <string>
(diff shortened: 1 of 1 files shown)


xhochy:
*bump* can any maintainer have a look at this?

src/stl_logging_unittest.cc
```diff
@@ -33,6 +33,7 @@
 
 #include <functional>
 #include <iostream>
+#include <iterator>
```
sergiud: `std::find` requires `<algorithm>`. What do you need `<iterator>` for?

PR #460, opened 2019-07-26
googletest: Switch to nanosleep

neheb:
usleep is deprecated and optionally not available with uClibc-ng.

### DIFF
--- a/src/googletest.h
+++ b/src/googletest.h
@@ -574,7 +574,12 @@ class Thread {
 
 static inline void SleepForMilliseconds(int t) {
 #ifndef OS_WINDOWS
+# if defined(_POSIX_C_SOURCE) && _POSIX_C_SOURCE >= 199309L
+  const struct timespec req = {0, t * 1000 * 1000};
+  nanosleep(&req, NULL);
+# else
   usleep(t * 1000);
+# endif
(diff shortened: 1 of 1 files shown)


src/googletest.h
```diff
@@ -574,7 +574,8 @@ class Thread {
 
 static inline void SleepForMilliseconds(int t) {
 #ifndef OS_WINDOWS
-  usleep(t * 1000);
+  const struct timespec req = {0, t * 1000 * 1000};
+  nanosleep(&req, NULL);
```
sergiud: Don't you have to check for `_POSIX_C_SOURCE >= 199309L`?

By the way, this change should be particularly submitted to the upstream (if not done yet).

neheb: What upstream?

neheb: If you mean googletest, it's already using nanosleep: https://github.com/google/googletest/blob/master/googletest/include/gtest/internal/gtest-port.h#L1138

(strange as C++11 has sleep_for for exactly this purpose).

neheb: wait a minute...even this is C++.

neheb:
Added check for _POSIX_C_SOURCE

PR #468, opened 2019-09-05
Fix const_basename on windows host and linux target

zaporozhets:
I use GCC for arm linux target compiled for windows host(Linaro arm-linux-gnueabihf for mingw). And const_basename doesn't work. On my linux taget I see full path instead of basename. e.g.

I0905 19:46:34.622531  1536 Z:\dev\visa_server\src\main.cpp:115] Hello World
I0905 19:46:34.628554  1536 Z:\dev\visa_server\src\main.cpp:123] Yo

### DIFF
--- a/src/utilities.cc
+++ b/src/utilities.cc
@@ -284,10 +284,8 @@ pid_t GetTID() {
 
 const char* const_basename(const char* filepath) {
   const char* base = strrchr(filepath, '/');
-#ifdef OS_WINDOWS  // Look for either path separator in Windows
   if (!base)
     base = strrchr(filepath, '\\');
-#endif
   return base ? (base+1) : filepath;
 }
 


zaporozhets:
@sergiud, could you please take a look at this pull request. cheers.

src/utilities.cc
```diff
@@ -284,10 +284,8 @@ pid_t GetTID() {
 
 const char* const_basename(const char* filepath) {
   const char* base = strrchr(filepath, '/');
-#ifdef OS_WINDOWS  // Look for either path separator in Windows
   if (!base)
```
sergiud: Backslashes are very unlikely on systems other than Windows. One should better fix the `OS_WINDOWS` define or directly check for `__MINGW__`.

PR #478, opened 2019-10-31
googletest.h: Replace deprecated usleep for sleep_for

neheb:
sleep_for is more portable. On Linux, it calls nanosleep, which is not
deprecated.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

note that glog does not compile using a C++98 library so this should be fine.

### DIFF
--- a/src/googletest.h
+++ b/src/googletest.h
@@ -41,9 +41,11 @@
 #include <setjmp.h>
 #include <time.h>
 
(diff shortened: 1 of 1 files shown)


neheb:
Looks like I'm mistaken actuially. There's a small bug. But otherwise it compiles.

PR #485, opened 2019-11-02
src/symbolize.cc: do not check for HAVE_DLFCN_H for macOS

ffontaine:
This seems to raise a build failure, see
https://github.com/google/glog/commit/10498b485fe1007f8e913bf2c39ff39fe01c8dc8#commitcomment-35770392

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


Yannic:
Thanks! Looks like this fixed it: https://buildkite.com/bazel/google-logging/builds/761

ffontaine:
That's good news, don't hesitate to ping me again if one of my commits raise other issues on macOS.

Issue #472, opened 2019-10-17
Bazel support for windows


skeptic-monkey:
Using google/glog from a bazel project on windows doesn't work.
bazel/glog.bzl doesn't seem to be windows compatible (but does on linux and darwing). Is there a plan to add support for it ? 

here is the error I get : 
```
$> bazel build --compiler=clang-cl --cxxopt=/std:c++17 @com_google_glog//:glog
INFO: Writing tracer profile to 'C:/users/lucbe/_bazel_lucbe/ixjskkhv/command.profile.gz'
INFO: Analyzed target @com_google_glog//:glog (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: C:/users/lucbe/_bazel_lucbe/ixjskkhv/external/com_google_glog/BUILD:5:1: C++ compilation of rule '@com_google_glog//:glog' failed (Exit 1)
In file included from external/com_google_glog/src/demangle.cc:38:
external/com_google_glog/src/utilities.h(80,11): fatal error: 'port.h' file not found
# include "port.h"
          ^~~~~~~~
1 error generated.
Target @com_google_glog//:glog failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 1.512s, Critical Path: 1.05s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```

drigz:
I don't have plans to work on this, and I don't have much Windows dev experience. If someone wants to take it on, let me know and I'll happily provide guidance on the Bazel side of things.

The first step to resolving the `port.h` error would be to change `bazel/glog.bzl` to use [select()](https://docs.bazel.build/versions/master/be/functions.html#select) to use the files in `src/windows/` instead of the Linux/Darwin header files.

The PR that fixes the Windows build should also enable Windows CI by adding to `.bazelci/presubmit.yml`:

```
  windows:
    # Optional: use VS 2017 instead of 2015.
    environment:
      BAZEL_VC: "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\VC"
    build_targets:
    - "..."
    test_targets:
    - "..."
```

mehrdadn:
Hi, for Windows support I think you need to do the following:

1. Add `src/glog/log_severity.h` to `srcs`

1. Add `src/windows/config.h`, `src/windows/port.cc`, `src/windows/port.h` as Windows-only `srcs` 
[...]

PR #488, opened 2019-11-04
Add windows support for bazel build

skeptic-monkey:
Attempt to fix https://github.com/google/glog/issues/472

linux/darwnin builds are not strictly equivalent to windows build. But it seem to already be that case without bazel, as we use pre-generated glog headers, that have hardcoded namespace for example.

Also, it doesn't support static linkage, but again, that seemed to already be hardcoded that way for windows when using pre-generated headers.

Windows support is far from perfect but it is at a least a step forward.

### DIFF
--- a/bazel/glog.bzl
+++ b/bazel/glog.bzl
@@ -7,95 +7,95 @@
 #       https://github.com/google/glog/issues/61
 #       https://github.com/google/glog/files/393474/BUILD.txt
 
-def glog_library(namespace='google', with_gflags=1, **kwargs):
-    if native.repository_name() != '@':
-        gendir = '$(GENDIR)/external/' + native.repository_name().lstrip('@')
+def glog_library(namespace = "google", with_gflags = 1, **kwargs):
+    if native.repository_name() != "@":
+        gendir = "$(GENDIR)/external/" + native.repository_name().lstrip("@")
     else:
-        gendir = '$(GENDIR)'
+        gendir = "$(GENDIR)"
 
     native.cc_library(
-        name = 'glog',
-        visibility = [ '//visibility:public' ],
+        name = "glog",
+        visibility = ["//visibility:public"],
         srcs = [
-            ':config_h',
-            'src/base/commandlineflags.h',
-            'src/base/googleinit.h',
-            'src/base/mutex.h',
-            'src/demangle.cc',
-            'src/demangle.h',
-            'src/logging.cc',
-            'src/raw_logging.cc',
-            'src/signalhandler.cc',
-            'src/stacktrace.h',
-            'src/stacktrace_generic-inl.h',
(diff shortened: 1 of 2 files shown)


skeptic-monkey:
@googlebot I signed it!

bazel/glog.bzl
```diff
@@ -38,58 +78,65 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
...
+        name = "glog_headers",
+        deps = select({
+            "@bazel_tools//src/conditions:windows": [":windows_glog_headers"],
+            "//conditions:default": [":default_glog_headers"],
+        }),
+    )
+
+    native.cc_library(
+        name = "windows_glog_headers",
+        hdrs = ["src/glog/log_severity.h", "src/windows/config.h",] + native.glob(["src/windows/glog/*.h"]),
```
drigz: This will leak config.h into the include path, which may break targets that depend on glog and contain `#include "config.h"`. Can you apply the pattern from https://github.com/bazelbuild/bazel/issues/4463 to calculate the path of `src/windows/`, and add `"-I" + src_windows` to `windows_only_copts`?

bazel/glog.bzl
```diff
@@ -38,58 +78,65 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
...
+        includes = ["src/windows"],
+        # config.h for windows seem hardcoded that way,
+        # and we need to propagate those defines to binaries/libraries linking
+        # against glog.
+        defines = [
+            "GOOGLE_GLOG_IS_A_DLL=1",
+            "GOOGLE_GLOG_DLL_DECL=__declspec(dllexport)",
+            "GOOGLE_GLOG_DLL_DECL_FOR_UNITTEST=__declspec(dllimport)",
```
drigz: Are these necessary in `defines`? If so, are they all necessary or can some be in `copts`?

It looks to me like the public headers have an `#ifndef GOOGLE_GLOG_DLL_DECL` section that should make them unnecessary, but perhaps I'm missing something.

drigz:
Hi @skeptic-monkey, will you have bandwidth to take a look at the review comments? If not, just let me know, as this is a valuable change as-is, so I'd be happy to merge it and take a look at the `config.h` issue myself.

skeptic-monkey:
I'll be able to have a look this weekend.  From what I've seen after checking quickly,  It's not that straightforward.   

Anyway. I'll check and if I don't see an easy solution I'll let you know

drigz:
You were right, it wasn't straightfoward :) But I managed to find a workaround, so I took the liberty of merging this and creating a PR for the cleanups.

Thanks for the contribution!

shahms:
FWIW, this breaks clients that use `features = ["layering_check"]` because the public glog library no longer exports any headers directly.

PR #499, opened 2019-11-20
Address review comments for Bazel/Win32 build

drigz:
This prevents leaking `config.h` and some unnecessary defines to rules that depend on glog. It also adds a caveat about not supporting the `namespace` parameter.

@skeptic-monkey, could you check that these changes work with your use case?


drigz:
Another user reported that this works for them, so I'll merge.

PR #502, opened 2019-12-01
src/logging.cc: encapsulate log cleaner, match logfile with <base_filename_><date>-<time>.<pid>

aesophor:
Previously log cleaner could be broken if users use `SetLogDestination()` to set a custom `base_filename_`, as discussed here: https://github.com/google/glog/pull/432#issuecomment-560052270

This PR fixes this bug by checking if a log's filename starts with `base_filename_`.


sergiud:
@cswuyg Could you please review the changes?

aesophor:
@sergiud I've been using this patch in my own projects for a week and it seems to work well. Tested on both Linux and Windows.

aesophor:
Please don't merge it yet, I've found a potential bug:

Currently I'm checking if a file is a glog's logfile by **checking if its filename starts with base_filename_**, but this might also delete non-glog logfiles. I should make sure the entire filename matches the format of a glog's filename.

aesophor:
@sergiud I've revised the patch and re-tested it. `LogCleaner::IsLogFromCurrentProject()` will now check if the given filename matches the format `<base_filename><YYYYMMDD>-<HHMMSS>.<pid>`.

sergiud:
Please resolve the conflict.

aesophor:
@sergiud I've resolved the conflict. Thanks!

PR #513, opened 2020-01-07
Use file offsets in traces if object has no symbols

mspang:
Use file offsets in traces if object has no symbols

Fallback to using file offsets if no symbol is found, like we do if the
object could not be opened.

This makes backtraces usable even if objects in the trace are stripped,
since the trace can be symbolized by post-processing it with a tool like
asan_symbolize.py.

Note that this is not currently compatible with SymbolizeCallback as
this overwrites the filename in the buffer. The behavior is unchanged
in that case.

Closes: #514

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ option (WITH_THREADS "Enable multithreading support" ON)
 option (WITH_TLS "Enable Thread Local Storage (TLS) support" ON)
 option (BUILD_SHARED_LIBS "Build shared libraries" OFF)
 option (PRINT_UNSYMBOLIZED_STACK_TRACES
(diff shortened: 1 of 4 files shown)


shinh:
Seems reasonable! Is there any reason someone want to turn off this feature? I'm not sure if we need `SYMBOLIZE_FALLBACK`.

mspang:
> Seems reasonable! Is there any reason someone want to turn off this feature? I'm not sure if we need `SYMBOLIZE_FALLBACK`.

I agree it's probably unnecessary; removed. Only added that in case someone is depending on the return value for something as we're unlikely to ever return "false" after this change. Personally I think
that is a feature :)

PR #517, opened 2020-01-16
Add vcpkg installation instructions

LilyWangL:
Glog is available as a port in vcpkg, a C++ library manager that simplifies installation for glog and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build glog, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and here is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -76,7 +76,7 @@ You can download and install glog using the [vcpkg](https://github.com/Microsoft
     cd vcpkg
     ./bootstrap-vcpkg.sh
     ./vcpkg integrate install
-    vcpkg install glog
+    ./vcpkg install glog
 
 The glog port in vcpkg is kept up to date by Microsoft team members and community contributors. If the version is out of date, please [create an issue or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.
 
--- a/INSTALL
+++ b/INSTALL
@@ -67,20 +67,6 @@ frame pointer based stack unwinder by passing the
 --enable-frame-pointers flag to configure.
 
 
-Building glog - Using vcpkg
-===========================
-
-You can download and install glog using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
-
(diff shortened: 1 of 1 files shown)


LilyWangL:
> @googlebot I signed it!

INSTALL
```diff
@@ -67,6 +67,20 @@ frame pointer based stack unwinder by passing the
 --enable-frame-pointers flag to configure.
 
 
+Building glog - Using vcpkg
```
sergiud: Please place the instructions after the basic installation section.

INSTALL
```diff
@@ -67,6 +67,20 @@ frame pointer based stack unwinder by passing the
...
+===========================
+
+You can download and install glog using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
```
sergiud: This is not a markdown file, so such links will not work here.

INSTALL
```diff
@@ -67,6 +67,20 @@ frame pointer based stack unwinder by passing the
...
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    ./vcpkg install glog
+
+The glog port in vcpkg is kept up to date by Microsoft team members and community contributors. If the version is out of date, please [create an issue or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.
```
sergiud: The sentence exceeds the maximum number of characters used in this file. Either format the sentence accordingly or convert the file to markdown. The latter would fix the issues with your links.

INSTALL
LilyWangL: Done.

PR #518, opened 2020-01-17
Add target existence checks to Unwind find module

MichaelDarr:
The build systems of projects depending on Glog may call the Unwind find
module multiple times. In these cases, the current unwind find module tries
to create a duplicate unwind::unwind target, crashing the build. This
patch adds an existence check before target creation to fix this issue.

Signed-off-by: Michael Darr <mdarr@matician.com>

### DIFF
--- a/cmake/FindUnwind.cmake
+++ b/cmake/FindUnwind.cmake
@@ -62,15 +62,17 @@ find_package_handle_standard_args (Unwind REQUIRED_VARS Unwind_INCLUDE_DIR
   Unwind_LIBRARY Unwind_PLATFORM_LIBRARY VERSION_VAR Unwind_VERSION)
 
 if (Unwind_FOUND)
-  add_library (unwind::unwind INTERFACE IMPORTED)
+  if (NOT TARGET unwind::unwind)
+    add_library (unwind::unwind INTERFACE IMPORTED)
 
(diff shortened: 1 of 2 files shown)


cmake/FindUnwind.cmake
```diff
@@ -62,15 +62,17 @@ find_package_handle_standard_args (Unwind REQUIRED_VARS Unwind_INCLUDE_DIR
...
+    set_property (TARGET unwind::unwind PROPERTY
+      INTERFACE_INCLUDE_DIRECTORIES ${Unwind_INCLUDE_DIR}
+    )
+    set_property (TARGET unwind::unwind PROPERTY
+      INTERFACE_LINK_LIBRARIES ${Unwind_LIBRARY} ${Unwind_PLATFORM_LIBRARY}
+    )
+    set_property (TARGET unwind::unwind PROPERTY
+      IMPORTED_CONFIGURATIONS RELEASE
+    )
+  endif(NOT TARGET unwind::unwind)
```
sergiud: Nitpick: missing space after `endif`.

MichaelDarr: Fixed! ad9505598d749f7084ad51fa5ee0ae9c4771c3d6

PR #523, opened 2020-02-06
Allow CMake user to disable Symbolize functionality

edbaunton:
Currently the user has to rely on the HAVE_SYMBOLIZE detection inside
CMakeLists.txt without having any control over whether it should be used.

Add support for disabling HAVE_SYMBOLIZE at configure time so user can specify
`cmake -DHAVE_SYMBOLIZE=0`.

Signed-off-by: Ed Baunton <ebaunton1@bloomberg.net>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,7 @@ option (PRINT_UNSYMBOLIZED_STACK_TRACES
   "Print file offsets in traces instead of symbolizing" OFF)
 option (WITH_PKGCONFIG "Enable pkg-config support" ON)
 option (WITH_UNWIND "Enable libunwind support" ON)
+option (WITH_SYMBOLIZE "Enable symbolize module" ON)
 
 if (NOT WITH_UNWIND)
   set (CMAKE_DISABLE_FIND_PACKAGE_Unwind ON)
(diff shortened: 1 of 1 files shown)


edbaunton:
On the platform I am testing on which doesn't have dladdr functionality, I think I am falling foul of [this](https://github.com/google/glog/blob/195d416e3b1c8dc06980439f6acd3ebd40b6b820/CMakeLists.txt#L403-L405) check which I think could simply be rewritten as:

`elseif (HAVE_DLADDR)`

because the [CMake Documentation](https://cmake.org/cmake/help/v3.6/variable/UNIX.html) reports that `UNIX` will return True on Apple.

However I am not familiar enough to know whether that logic is like that for some hidden reason.

CMakeLists.txt
```diff
@@ -365,44 +365,46 @@ if (HAVE_EXECINFO_H)
   set (HAVE_STACKTRACE 1)
 endif (HAVE_EXECINFO_H)
 
-if (WIN32 OR CYGWIN)
-  cmake_push_check_state (RESET)
-  set (CMAKE_REQUIRED_LIBRARIES DbgHelp)
+if(NOT DEFINED HAVE_SYMBOLIZE)
```
sergiud: Please insert spaces after control statements and function names.

CMakeLists.txt
```diff
@@ -365,44 +365,46 @@ if (HAVE_EXECINFO_H)
...
 
   cmake_pop_check_state ()
 
-  if (HAVE_SYMBOLIZE)
-    set (HAVE_STACKTRACE 1)
-  endif (HAVE_SYMBOLIZE)
-elseif (UNIX OR (APPLE AND HAVE_DLADDR))
-  set (HAVE_SYMBOLIZE 1)
-endif (WIN32 OR CYGWIN)
+    if (HAVE_SYMBOLIZE)
```
sergiud: The correct way of doing is to add an `WITH_` option. `HAVE_*` variables are not intended to be modified by users.

edbaunton: The implementation has been updated to wrap the block with `WITH_SYMBOLIZE` but still use `HAVE_SYMBOLIZE` internally.

edbaunton:
Made the changes.

PR #526, opened 2020-02-16
fix bazel build for freebsd

dfr:
The installed version of unwind.h on FreeBSD doesn't declare _Unwind_Backtrace and related types unless _GNU_SOURCE is defined.


dfr:
Thanks!

PR #529, opened 2020-03-06
Support stackstrace for aarch64

huangtianhua:


huangtianhua:
@googlebot I signed it!

src/stacktrace_aarch64-inl.h
```diff
@@ -0,0 +1,105 @@
...
+//
+// Author: Craig Silverstein
+//
+// Produce stack trace.  I'm guessing (hoping!) the code is much like
+// for x86.  For apple machines, at least, it seems to be; see
+//    http://developer.apple.com/documentation/mac/runtimehtml/RTArch-59.html
+//    http://www.linux-foundation.org/spec/ELF/ppc64/PPC-elf64abi-1.9.html#STACK
+// Linux has similar code: http://patchwork.ozlabs.org/linuxppc/patch?id=8882
+
+#include <stdio.h>
```
sergiud: Please use `<cstdio>`.

src/stacktrace_aarch64-inl.h
```diff
@@ -0,0 +1,105 @@
...
+#include <stdint.h>   // for uintptr_t
```
sergiud: One should include `<cstdint>`. However, the header is available only starting with C++11. Glog still uses C++03.

bazel/glog.bzl
```diff
@@ -83,6 +83,7 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
             "src/stacktrace_generic-inl.h",
             "src/stacktrace_libunwind-inl.h",
             "src/stacktrace_powerpc-inl.h",
+	    "src/stacktrace_aarch64-inl.h",
```
sergiud: Nitpick: inconsistent indentation.

huangtianhua: Done.

src/stacktrace_aarch64-inl.h
huangtianhua: I find other files use <stdio.h>?

src/stacktrace_aarch64-inl.h
huangtianhua: same

huangtianhua:
@sergiud , Thanks for reviewing. And I have a question about getting stacktrace by libunwind, if HAVE_LIB_UNWIND is set(like the case apache/kudu using), we will get stack trace by libunwind before my commit, but seems the function 'GetStackTrace' of stacktrace_libunwind-inl.h dosen't work on aarch64, so I changed the logic to use stacktrace_aarch64-inl.h(which I newly added) if HAVE_LIB_UNWIND is set. So I wonder it is the right way to get stack trace by libunwind on aarch64?  or it doesn't support on aarch64? or how to modify the file stacktrace_libunwind-inl.h to support get stack trace on aarch64? Thanks very much.

src/stacktrace_aarch64-inl.h
sergiud: Ok then.

src/stacktrace_aarch64-inl.h
sergiud: Not quite. You need a fallback to C++03 where the header is not available.

huangtianhua: Other files include <stdint.h>, so I think it works.

sergiud: Including is okay, but you still need a fallback.

romange:
@huangtianhua would it be possible to finish this PR?

sergiud:
@romange Could you please test the changes?

sergiud (approved):

romange:
@sergiud  The bug in libunwind seems to be fixed https://github.com/libunwind/libunwind/issues/235  

Do you still want to commit the workaround for aarch64 or you prefer to wait for the next version of libunwind to be released?

sergiud:
@romange Thanks for the info. I'd rather avoid the workaround.

PR #545, opened 2020-04-27
OpenBSD/powerpc: clang does not define _CALL_*, use proper ifdef instead

julianaito:
Hi,

OpenBSD/macppc uses now clang by default and does not define _CALL_SYSV,
it's needed to use the already existing alternative for such platforms.


julianaito:
@googlebot I signed it!

PR #546, opened 2020-04-27
Fix build for Emscripten

schoppmp:
The current Bazel build configuration assumes that <sys/syscall.h> is present on non-windows targets, which is not the case when building with [emscripten](https://github.com/emscripten-core/emscripten). This PR fixes the issue.

### DIFF
--- a/bazel/glog.bzl
+++ b/bazel/glog.bzl
@@ -48,6 +48,28 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
         "-I%s/glog_internal" % gendir,
     ]
 
+    native.config_setting(
+        name = "wasm",
+        values = {"cpu": "wasm"},
+    )
+
+    wasm_copts = [
+        # Inject a C++ namespace.
+        "-DGOOGLE_NAMESPACE='%s'" % namespace,
+        # Allows src/base/mutex.h to include pthread.h.
+        "-DHAVE_PTHREAD",
+        # Allows src/logging.cc to determine the host name.
+        "-DHAVE_SYS_UTSNAME_H",
+        # For src/utilities.cc. We have time.h, but not syscall.h.
+        "-DHAVE_SYS_TIME_H",
+        # Enable dumping stacktrace upon sigaction.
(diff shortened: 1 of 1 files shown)


tuankiet65:
Seems like this is only fixed for Bazel but not CMake. I have a CMake project and while compiling I got the error that syscall is not declared:
```
In file included from third_party/glog/src/raw_logging.cc:34:
In file included from third_party/glog/src/utilities.h:84:
build/emscripten-debug/third_party/glog/glog/logging.h:1438:68: warning: unused parameter 'usecs' [-Wunused-parameter]
                    const char* message, size_t message_len, int32 usecs) {
                                                                   ^
third_party/glog/src/raw_logging.cc:139:3: error: use of undeclared identifier 'syscall'
  safe_write(STDERR_FILENO, buffer, strlen(buffer));
  ^
third_party/glog/src/raw_logging.cc:63:34: note: expanded from macro 'safe_write'
# define safe_write(fd, s, len)  syscall(SYS_write, fd, s, len)
                                 ^
1 warning and 1 error generated.
```

bazel/glog.bzl
```diff
@@ -18,14 +18,20 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
         gendir = "$(GENDIR)"
         src_windows = "src/windows"
 
+    # Config setting for WebAssembly target.
+    native.config_setting(
+        name = "wasm",
+        values = {"cpu": "wasm"},
```
attilaolah, line 24: Shortly before this code was submitted, `wasm32` and `wasm64` were added in https://github.com/bazelbuild/platforms/commit/966defdc808b7bd3c966121131eb07d9668f5ff9 as standard WebAssembly CPU architectures: [BUILD#L83-91](https://github.com/bazelbuild/platforms/blob/aee75f4429c6ea769c60f1aeaa5d078cffcd5c8d/cpu/BUILD#L83-L91).

Maybe this should be changed accordingly (i.e. split into two config settings, and list them both in the `select` below).

PR #559, opened 2020-06-19
Remove dead references to bluelist/blacklist classes

DigiTec:
This propagates the term blacklist which holds negative racial connotations across a large set of projects. I initially came to replace this term, but I've now noticed it seems to be unused.


sergiud:
I would kindly ask you to remove the patronizing language from you commit message and only state the technical reasons for your changes. You are in no position to decide whether the term in question hold negative connotations. In fact, it is the first time I hear that the technical term "blacklist" is now offensive.

Please do not understand my change request as a reason for starting a political discussion. In contrary, I ask you limit any discourse solely to technical issues. Thank you.

sergiud:
Obsoleted by #585.

PR #580, opened 2020-09-16
Make cmake build on path with spaces.

Yicong-Huang:
wrap the path with double quotation so that it can build on path with spaces.


Yicong-Huang:
@googlebot I signed it!

PR #587, opened 2020-10-06
Implemented GitHub Actions support

leozz37:
### GitHub Actions

A simple CI for building Bazel. There's also support for actions/cache, making the builds much faster (least the first time).

A successfully run can be seen [here](https://github.com/leozz37/glog/actions).


drigz:
Thanks for the PR and apologies for not responding sooner.

We already have Buildkite CI enabled (although admittedly the status isn't showing up on PRs https://github.com/bazelbuild/continuous-integration/issues/1095). Since it also tests Mac/Windows and warns us about breakage for future Bazel versions, I'm inclined to stick with it.

Assuming we can fix the presubmit issue, would you suggest that we close this PR, or is there a reason to use GitHub Actions in addition?

Issue #599, opened 2020-12-17
No system link created along with the log files


zoukyle:
How to reproduce:
Use the bazel environment:
```
################################################################################
# Define external binding for GFlags : https://github.com/gflags
# Use: //external:gflags
################################################################################
http_archive(
    name = "com_github_gflags_gflags",
    sha256 = "34af2f15cf7367513b352bdcd2493ab14ce43692d2dcd9dfc499492966c64dcf",
    strip_prefix = "gflags-2.2.2",
    urls = ["https://github.com/gflags/gflags/archive/v2.2.2.tar.gz"],
)

bind(
    name = "gflags",
    actual = "@com_github_gflags_gflags//:gflags",
)

################################################################################
# Define external binding for glog: https://github.com/google/glog
# Use: //external:glog
################################################################################
http_archive(
    name = "com_github_google_glog",
    sha256 = "62efeb57ff70db9ea2129a16d0f908941e355d09d6d83c9f7b18557c0a7ab59e",
    strip_prefix = "glog-d516278b1cd33cd148e8989aec488b6049a4ca0b",
    urls = ["https://github.com/google/glog/archive/d516278b1cd33cd148e8989aec488b6049a4ca0b.zip"],
)

bind(
    name = "glog",
    actual = "@com_github_google_glog//:glog",
)

The cc main file:
int main(int argc, char* argv[]) {
  google::InitGoogleLogging(argv[0]);
  google::InstallFailureSignalHandler();

  LOG(INFO) << "Initialized google logging";
  LOG(WARNING) << "This is a warning";
  LOG(ERROR) << "This is an error";

  CHECK(true);
  CHECK_GT(2, 1);
  CHECK_LE(1, 1);

  LogIfGt(2, 1);
  LogIfGt(1, 2);
  for (int i = 0; i < 200; ++i) {
    LOG_EVERY_N(INFO, 100) << "Log every 100, i is " << i;
  }

  return 0;
}
```

Run the example with: GLOG_log_dir=./tmp/ GLOG_alsologtostderr=1 bazel-bin/experimentals/examples/glog/example
The log dir has:
example.liang.liangzou.log.ERROR.20201217-090344.91782  example.liang.liangzou.log.INFO.20201217-090344.91782  example.liang.liangzou.log.WARNING.20201217-090344.91782
But there are no example.INFO etc system links created.

PR #600, opened 2020-12-17
Fix the symlink creation bug if using bazel

zoukyle:
This will fix https://github.com/google/glog/issues/599

### DIFF
--- a/bazel/glog.bzl
+++ b/bazel/glog.bzl
@@ -56,6 +56,8 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
     linux_or_darwin_copts = wasm_copts + [
         # For src/utilities.cc.
         "-DHAVE_SYS_SYSCALL_H",
+        # For src/logging.cc to create symlinks.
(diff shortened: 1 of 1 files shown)


zoukyle:
@googlebot I signed it!

bazel/glog.bzl
```diff
@@ -56,6 +56,7 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
     linux_or_darwin_copts = wasm_copts + [
         # For src/utilities.cc.
         "-DHAVE_SYS_SYSCALL_H",
+        "-DHAVE_UNISTD_H",
```
drigz, line 59: Could you add a comment `# For -log_link.`? `For src/utilities.cc.` is not so relevant here.

drigz (changes requested):
Thanks!

bazel/glog.bzl
zoukyle: Done

drigz (approved):

PR #614, opened 2021-03-01
Rename BUILD to BUILD.bazel.

### LINKED ISSUE #606: BUILD file will create a conflict with build/ folder
Hi,

In the repo there is a [BUILD file](https://github.com/google/glog/blob/master/BUILD) that creates a conflict when following the [CMake instructions](https://github.com/google/glog/blob/master/README.rst#cmake) because it will try to create a `build/` folder.

I renamed the BUILD file and then I've been able to make the project.

drigz:
This avoids conflicts with build/ on case-insensitive filesystems.

Fixes #606.


drigz:
Manually tested that this doesn't interfere with the [build_file](https://docs.bazel.build/versions/master/repo/http.html#http_archive-build_file) attribute. I can't think of any other way downstream users might depend on the name of this file, so hopefully this is safe.

PR #620, opened 2021-03-30
cmake: allow to disable symbolize

### LINKED ISSUE #563: Disabling 'Symbolize functionality' is not working
From #523 
`cmake -DWITH_SYMBOLIZE=OFF` is not working, because `HAVE_SYMBOLIZE` is defined in:
https://github.com/google/glog/blob/0a2e5931bd5ff22fd3bf8999eb8ce776f159cda6/src/utilities.h#L130-L141

sergiud:
Closes #563

@drigz ~~This change might break Bazel builds. Maybe there is an alternative to removal?~~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -416,7 +416,7 @@ if (WITH_SYMBOLIZE)
     }
     ]=] HAVE_SYMBOLIZE)
 
-  cmake_pop_check_state ()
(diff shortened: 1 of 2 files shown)


sergiud:
@drigz I reverted the breaking changes. Bazel builds are now unaffected.

drigz (approved):
Thanks!

PR #628, opened 2021-03-31
bazel: use expand_template instead of genrules

drigz:
This lets us drop the dependency on the preprocessed headers in
src/windows/glog/.

For some reason, this lets Bazel realize we've failed to declare a
dependency on dirent.h, so I added that too.

### DIFF
--- a/bazel/glog.bzl
+++ b/bazel/glog.bzl
@@ -9,6 +9,28 @@
 #
 # Known issue: the namespace parameter is not supported on Win32.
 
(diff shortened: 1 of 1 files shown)


drigz:
@sergiud This should unblock removal of the preprocessed headers. I'm not expecting a detailed review, but maybe you could rubber stamp this if it looks mostly OK to you :)

sergiud (approved):
@drigz Thanks! I skimmed through the changes and they look good to me!

PR #635, opened 2021-04-06
Use generated config.h for Windows.

drigz:
This unblocks a simplification of the Windows CMake build.


sergiud (approved):
Works great, thanks!

PR #639, opened 2021-04-08
fixed windows builds

sergiud:


bazel/glog.bzl
```diff
@@ -99,7 +99,6 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
 
     windows_only_srcs = [
         "src/glog/log_severity.h",
-        "src/windows/config.h",
```
sergiud, line 102: @drigz It seems I overlooked this last dependency to `windows/config.h`. This should not be necessary anymore, correct? Unfortunately, Bazel build fails without the Windows header.

drigz: https://github.com/google/glog/commit/30213bf96b9ade105d149bb010332923f11de920

You might need to fold it into this PR as it requires simultaneous deletion of the file.

sergiud: Great, thank you. I had to additionally define `GLOG_NO_ABBREVIATED_SEVERITIES` for Bazel builds on Windows. I hope my change is acceptable.

bazel/glog.bzl
```diff
@@ -93,13 +93,13 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
     ]
 
     windows_only_copts = [
+        "-DGLOG_NO_ABBREVIATED_SEVERITIES",
```
drigz, line 96: This seems like it could break users. Can you add a #define to port.h instead, along the lines of the docs? https://github.com/google/glog#notes-for-windows-users

drigz (changes requested):

bazel/glog.bzl
drigz: On second thoughts, as it's in copts rather than defines, it's unlike to affect users but it would still be cleaner to use the approach in the docs IMO.

sergiud: Are you saying that the define is transitive, i.e., it is forwarded to library's consumer? Either way, CMake always defines `GLOG_NO_ABBREVIATED_SEVERITIES` on Windows to avoid conflicts due to `ERROR`:

https://github.com/google/glog/blob/a79416bab739907a31a1e139540921ff0492f010/CMakeLists.txt#L598-L600

If I recall correctly, on Windows glog without `GLOG_NO_ABBREVIATED_SEVERITIES` is unusable anyway. Regardless, we should probably sync the approaches taken by both build systems.

drigz: defines are transitive (useful when needed in public headers) but copts are not. If CMake does it too then this LGTM, thanks!

drigz (approved):

PR #641, opened 2021-04-08
CHECK support for nullptr (fixes #341)

### LINKED ISSUE #341: CHECK_XX variants do not compile if nullptr is an argument
This prevents things like:

CHECK_NE(ptr, nullptr);

from compiling

### LINKED ISSUE #562: compile error: use of overloaded operator '<<' is ambiguous
I'm using Ubuntu 18.04.4 LTS (Bionic Beaver), clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final), cmake version 3.10.2.
I cloned, compiled, installed the project, and trying to use it in my own project.
But once I included the header "glog/logging.h", I got an compile error like:
```
_/usr/local/include/glog/logging.h:649:9: error: use of overloaded operator '<<' is ambiguous 
(with operand types 'std::ostream' (aka 'basic_ostream<char>') and 'const nullptr_t')
  (*os) << v;_
```
Did this because the use of the operator is really ambiguous or there's implicit incompatible with some cmake settings I made?

____________________________________________________________________________________________

I found it is because that **CHECK_NE() is not suitable to estimate a pointer to nullptr.**
Try following code:
```c++
#include "glog/logging.h"
#include <iostream>

int main() {
    int i = 0;
    CHECK_NE(&i, nullptr);
    return 0;
}
```

sergiud:
This allows `CHECK_NE(foo, nullptr)` to compile and produces "nullptr" for the string representation of `nullptr`.

Finalizes #342

Closes #562 #341

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,6 +240,12 @@ std::atomic<int> i;
 int main() { }
 " HAVE_CXX11_ATOMIC)
 
+check_cxx_source_compiles ("
+#include <cstddef>
+void foo(std::nullptr_t) {}
+int main(void) { foo(nullptr); }
+" HAVE_CXX11_NULLPTR_T)
+
 if (WITH_TLS)
   # Cygwin does not support the thread attribute. Don't bother.
   if (HAVE_GCC_TLS)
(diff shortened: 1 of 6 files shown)


drigz (changes requested):
Thanks! Could you also add -DHAVE_CXX11_NULLPTR_T to common_copts in bazel/glog.bzl?

sergiud:
@drigz In that case, `ac_cv_cxx11_nullptr_t` should also be set to 1 in `common_config`, correct?

drigz:
> @drigz In that case, ac_cv_cxx11_nullptr_t should also be set to 1 in common_config, correct?

Ah, I misread and thought that it was set to 1, sorry. Either approach is OK, in that case, but I have a slight preference for assuming a modern environment in the Bazel build: I think that people with weird environments use build systems with environment detection, and Bazel users are expecting the modern features.

drigz (approved):
Thanks!

PR #650, opened 2021-05-10
Allow updating vmodule levels after vmodule level has been cached

### LINKED ISSUE #649: SetVLOGLevel does not work if VLOG_IS_ON has already cached its module level
below is the minimal example that demonstrates the issue. 
I can send the possible fix.

```
int TestVlogHelper() {
  if (VLOG_IS_ON(1)) {
    return 1;
  }
  return 0;
}

void TestVLogModule() {
  int c = TestVlogHelper();
  CHECK_EQ(0, c);

  CHECK_EQ(0, SetVLOGLevel("logging_unittest", 1));
  c = TestVlogHelper();   
  CHECK_EQ(1, c);   // Fails here.
}
```

romange:
We maintain a linked list of all sites so we could invalidate those that should change upon SetVLOGLevel call.
Once site level starts pointing to module override it will be removed from the linked list to avoid unnecessary traversals.

Fixes #649

### DIFF
--- a/src/logging_unittest.cc
+++ b/src/logging_unittest.cc
@@ -98,6 +98,7 @@ static void TestLogging(bool check_counts);
 static void TestRawLogging();
 static void LogWithLevels(int v, int severity, bool err, bool alsoerr);
 static void TestLoggingLevels();
+static void TestVLogModule();
 static void TestLogString();
 static void TestLogSink();
 static void TestLogToString();
@@ -223,6 +224,7 @@ int main(int argc, char **argv) {
   TestLogging(true);
   TestRawLogging();
   TestLoggingLevels();
+  TestVLogModule();
   TestLogString();
   TestLogSink();
   TestLogToString();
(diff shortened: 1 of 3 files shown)


romange:
@sergiud  I can not see why windows build is failing and I do not have a windows machine to test it. Could you help pointing out the problem?

sergiud:
The test fails because `TestVlogHelper` returns 0 instead of 1. Looks like `VLOG_IS_ON` does not work as expected anymore.

src/logging_unittest.cc
```diff
@@ -449,6 +451,22 @@ void TestLoggingLevels() {
...
+  return 0;
+}
+
+void TestVLogModule() {
+  int c = TestVlogHelper();
+  EXPECT_EQ(0, c);
+
+  EXPECT_EQ(0, SetVLOGLevel("logging_unittest", 1));
+  c = TestVlogHelper();
+  EXPECT_EQ(1, c);
```
sergiud, line 467: The test fails on this line.

sergiud: Likely related to this:

https://github.com/google/glog/blob/0b3d4cb471e36d8e677e75e4e8d0345e7bf2a418/src/glog/vlog_is_on.h.in#L90-L92

romange: I removed the test from non __GNUC__ environments but it still fails for mingw.

romange: I fixed the tests for mingw environment as well

romange:
@sergiud  it should pass now. PTAL

sergiud:
Thanks!

Issue #125, opened 2016-09-22
Allow glog to be queried as to whether or not InitGoogleLogging has been called.


KevinLucidyne:
If you call InitGoogleLogging twice, then you get a fatal error and the program ends.

However, there is no way to query glog as to whether glog has already been initialized or not.  This query is in the glog code (obviously) but it's not exposed in the API.  Please make this public.

Here is the use case where this is useful, for me at least.

I'm adding an unmanaged C++ package that uses glog to a managed C# DLL that is in turn dynamically loaded to a C# managed code application but I need to change the glog default behavior.  I can't call InitGoogleLogging in the application so instead I have to do the initialization in a nested singleton class in the DLL.

konstan:
+1!

debamax:
+1!

OasisArtisan:
This is still an issue but with a fairly easy work around. (Unless I'm missing something obvious)

We just need access to 
`bool IsGoogleLoggingInitialized();`
Which is defined in the `glog/src/utilities.h`

So just define the following in your own code.
```
namespace google {
  namespace glog_internal_namespace_ {
    extern bool IsGoogleLoggingInitialized();
  }
  const auto& IsGoogleLoggingInitialized = glog_internal_namespace_::IsGoogleLoggingInitialized;
}
```
The linker should take care of linking the symbol to its implementation in glog.

So now you can simply use the function as `google::IsGoogleLoggingInitialized()`

Cheers.

xkszltl:
@sergiud 
So how's the issue closed in 0.5?
Even worse, seems the `::google::glog_internal_namespace_::IsGoogleLoggingInitialized()` is now a hidden symbol, the redecl  hack won't work any more.

sergiud:
The issue received no activity, hence I closed it. You're welcome provide a PR.

xkszltl:
I think this issue is not about complaining change of internal API, but about why it is internal, because without it there's simply no reliable way to init without full control of process lifespan.
How about we simply alias it to a public API?

sergiud:
Sure, go ahead.

PR #651, opened 2021-05-13
Expose `IsGoogleLoggingInitialized()` in public API.

xkszltl:
Usually library does not have control of the process lifespan.
Without this function, it is impossible to init/shutdown reliably.
It has been one of the major pain points for years when using glog in libraries.

AFAIK 3 workarounds have been used previously:
1. Init without checking. This causes compatiblity issues with other libs using glog.
2. Also provide a init function in library's API. This makes API complicated and stateful, especially for libs that does not mean to stay for the entire life of process.
3. Steal the utility function in internal namespace. Does not work with msvc (due to missing dllexport) or `gcc -fvisibility=hidden`.

None of them are perfect, except for the last hack that usually works well on Linux.
0.5.0 changes default visibility to hidden and it does not work anymore.

Resolve https://github.com/google/glog/issues/125

### DIFF
--- a/src/utilities.cc
+++ b/src/utilities.cc
@@ -62,6 +62,10 @@ _START_GOOGLE_NAMESPACE_
 
 static const char* g_program_invocation_short_name = NULL;
 
+bool IsGoogleLoggingInitialized() {
+  return g_program_invocation_short_name != NULL;
+}
+
 _END_GOOGLE_NAMESPACE_
 
 // The following APIs are all internal.
@@ -174,10 +178,6 @@ const char* ProgramInvocationShortName() {
     // TODO(hamaji): Use /proc/self/cmdline and so?
     return "UNKNOWN";
   }
-}
-
-bool IsGoogleLoggingInitialized() {
-  return g_program_invocation_short_name != NULL;
 }
 
 #ifdef OS_WINDOWS
--- a/src/utilities.h
+++ b/src/utilities.h
@@ -163,8 +163,6 @@ namespace glog_internal_namespace_ {
 
 const char* ProgramInvocationShortName();
 
-bool IsGoogleLoggingInitialized();
-
 int64 CycleClock_Now();
 
 int64 UsecToCycles(int64 usec);
--- a/src/glog/logging.h.in
+++ b/src/glog/logging.h.in
@@ -594,6 +594,9 @@ GOOGLE_GLOG_DLL_DECL void InitGoogleLogging(const char* argv0,
                                             void* prefix_callback_data = NULL);
 #endif
 
+// Check if google's logging library has been initialized.
+GOOGLE_GLOG_DLL_DECL bool IsGoogleLoggingInitialized();
+
 // Shutdown google's logging library.
 GOOGLE_GLOG_DLL_DECL void ShutdownGoogleLogging();
 
--- a/src/logging_custom_prefix_unittest.cc
+++ b/src/logging_custom_prefix_unittest.cc
@@ -221,10 +221,14 @@ int main(int argc, char **argv) {
   LogWithLevels(0, 0, 0, 0);  // simulate "before global c-tors"
   const string early_stderr = GetCapturedTestStderr();
 
+  EXPECT_FALSE(IsGoogleLoggingInitialized());
+
   // Setting a custom prefix generator (it will use the default format so that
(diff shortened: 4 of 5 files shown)


xkszltl:
@sergiud

src/logging.cc
```diff
@@ -2568,6 +2568,10 @@ void InitGoogleLogging(const char* argv0,
 }
 #endif
 
+bool IsGoogleLoggingInitialized() {
+  return glog_internal_namespace_::IsGoogleLoggingInitialized();
```
sergiud, line 2572: The internal definition can be moved directly to the outer namespace. No aliasing necessary.

sergiud (changes requested):
A simple test checking the result of `IsGoogleLoggingInitialized` before and after init, and also after a shutdown should be added as well.

src/logging.cc
xkszltl: There're still possible ways to touch that symbol, e.g. when linking glog as static lib.
I don't want to break that backward compatibility unnecessarily.

xkszltl:
Tests added.

src/logging.cc
sergiud: Backwards compatibility for internal symbols does not matter,

xkszltl: I would say this one is a little bit special.
Because of its high importance (many use cases doesn't work without it), I believe there're many code out there touching this symbol regardless.
If we can easily keep the compatibility for free, why do we need to break it for nothing?

Let me know if I missed any strong benefit we can get by breaking it.

sergiud: The benefit is clean code. A change to an internal symbol does not justify maintaining compatibility that was not guaranteed in the first place.

xkszltl: > The benefit is clean code.

I can make the change if you think that's better.

> Your argument does not justify maintaining compatibility that was not guaranteed in the first place.

Exactly and intended.
IMHO when there's maintenance needed, it can be simply removed at that time as the cost comes.
Before that happens, there's be a free gap for graceful deprecation.

> In particular, the "breaking change" to the internal symbol has already occurred ...

Not for static build.
The change is only on visibility of shared libs.

xkszltl: Emmm, if I move it to logging.cc, I'll also need to expose the locally defined `g_program_invocation_short_name` itself probably into a header, or introduce another function to hide it that does basically the same as internal `IsGoogleLoggingInitialized`. Don't think that's a good idea.

sergiud: Please make the change.

xkszltl: Changed to exposing `g_program_invocation_short_name`.

src/utilities.cc
```diff
@@ -60,7 +60,7 @@ using std::string;
 
 _START_GOOGLE_NAMESPACE_
 
-static const char* g_program_invocation_short_name = NULL;
+const char* g_program_invocation_short_name = NULL;
```
sergiud, line 63: Removing `static` is a bad idea. Can't you just move the definition of `IsGoogleLoggingInitialized` a couple of lines to the top above `glog_internal_namespace_`? Only the declaration would need to be moved to `logging.h.in`.

src/utilities.h
```diff
@@ -149,6 +149,8 @@
 
 _START_GOOGLE_NAMESPACE_
 
+extern const char* g_program_invocation_short_name;
```
sergiud, line 152: Should not be necessary.

src/utilities.cc
xkszltl: Is it counter-intuitive to have decl in logging.h and def in utilities.cc, when both of them have their .h/.cc counterpart?

sergiud: Yes. However, the alternative would be to move all related functionality to `logging.cc` which I'd like to avoid at this point. For consistency, one can declare the function twice both in `logging.h.in` and `utilities.h`. But this is not really necessary.

xkszltl: Updated

src/utilities.h
xkszltl: Reverted

xkszltl:
Added decl in logging.h and def in utilities.h as requested.

sergiud (approved):

PR #654, opened 2021-05-18
Make LogCleaner support relative paths

### LINKED ISSUE #653: Crash if enable google::EnableLogCleaner(1);
I want to update glog-0.4.0 to glog-0.5.0,and wanna use Automatically Remove Old Logs functions.but my program crashed everytimes if enabled "google::EnableLogCleaner(1);"，Is there a problem with my usage?
（The static library is referenced in my program）

code attach:
```
// init log
const char *init_log(const char *cmd, const int log_level) {
    saas_uint8_t set_log_level = 0;
    FLAGS_logbufsecs = 0;
    google::InitGoogleLogging(cmd);

    if (check_log_level(log_level)) {
        set_log_level = log_level;
    }

//    google::InstallFailureSignalHandler();
      google::EnableLogCleaner(1);   // crash if uncomment.

    google::SetLogDestination(google::INFO, ".");
//    google::SetLogDestination(set_log_level, ".");
//    FLAGS_log_dir = "./logs";
	......
```

dump msg:
```
Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff6838921 in __GI_abort () at abort.c:79
#2  0x00007ffff722b957 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff7231ae6 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff7231b21 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x000055555578ffda in google::LogMessage::~LogMessage() ()
#6  0x00005555557648f2 in SConfig::read_config_from_file (this=0x7ffff4457080, filePath=0x5555564a476c "./SaasAccelerator.conf") at /home/workspace/src/SConfig.cpp:88
```

SConfig.cpp:88 :
```
 LOG(INFO) <<  "************start****************";
```

aesophor:
Closes #653 

In #653, the user reports that the following usage of glog API:
```cpp
google::SetLogDestination(google::INFO, ".");
google::EnableLogCleaner(1);
```

will lead to an uncaught out_of_range exception.
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::at: __n (which is 18446744073709551615) >= this->size() (which is 0)
[1]    835009 abort (core dumped)  ./out
```

This is caused by a bug in `LogCleaner::Run()` where `base_filename` doesn't end with a "/".
https://github.com/google/glog/blob/0b3d4cb471e36d8e677e75e4e8d0345e7bf2a418/src/logging.cc#L1334-L1335

This PR:
1. checks if `base_filename` in `LogCleaner::Run()` ends with a "/", and if not, then we simply pass the relative path to `LogCleaner::GetOverdueLogNames()`.
2. In `LogCleaner::GetOverdueLogNames`, we only concatenate `log_directory` and `ent->d_name` if `log_directory` ends with a "/". Otherwise, just pass the relative path to `LogCleaner::IsLogFromCurrentProject`.
3. Some trivial changes to improve the readability of source code

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -478,7 +478,6 @@ class LogFileObject : public base::Logger {
 class LogCleaner {
  public:
   LogCleaner();
-  virtual ~LogCleaner() {}
 
   void Enable(int overdue_days);
   void Disable();
@@ -486,7 +485,7 @@ class LogCleaner {
            const string& base_filename,
            const string& filename_extension) const;
 
-  inline bool enabled() const { return enabled_; }
+  bool enabled() const { return enabled_; }
 
  private:
   vector<string> GetOverdueLogNames(string log_directory,
@@ -1296,9 +1295,6 @@ void LogFileObject::Write(bool force_flush,
 
     // Remove old logs
     if (log_cleaner.enabled()) {
-      if (base_filename_selected_ && base_filename_.empty()) {
-        return;
-      }
       log_cleaner.Run(base_filename_selected_,
                       base_filename_,
                       filename_extension_);
@@ -1324,6 +1320,7 @@ void LogCleaner::Run(bool base_filename_selected,
                      const string& base_filename,
                      const string& filename_extension) const {
   assert(enabled_ && overdue_days_ >= 0);
+  assert(!base_filename_selected || !base_filename.empty());
 
   vector<string> dirs;
 
@@ -1361,7 +1358,6 @@ vector<string> LogCleaner::GetOverdueLogNames(string log_directory,
   // Try to get all files within log_directory.
   DIR *dir;
   struct dirent *ent;
-
 
   if ((dir = opendir(log_directory.c_str()))) {
     while ((ent = readdir(dir))) {


aesophor:
@sergiud @jixianliu1234

I've also discovered this:

```cpp
google::SetLogDestination(google::INFO, "./");  // creates logs under process's cwd
                                                // example: 20210518-122553.954758

google::SetLogDestination(google::INFO, ".");   // also creates logs under process's cwd, but filenames starts with "."
                                                // example: .20210518-122537.954584
```

Another example:
```cpp
google::SetLogDestination(google::INFO, "/tmp/");  // creates logs under /tmp/
google::SetLogDestination(google::INFO, "/tmp");   // creates logs under / with filenames starting with "tmp"
```

<br>

Therefore, I think the correct way of setting log destination to "." is `google::SetLogDestination(google::INFO, "./");`.
Maybe this should be documented somewhere (please tell me if I missed it in the documentation)

Thanks.

src/logging.cc
```diff
@@ -1359,17 +1362,18 @@ vector<string> LogCleaner::GetOverdueLogNames(string log_directory,
...
+  if ((dir = opendir(log_directory.c_str()))) {
+    while ((ent = readdir(dir))) {
       if (!strcmp(ent->d_name, ".") || !strcmp(ent->d_name, "..")) {
         continue;
       }
-      string filepath = log_directory + ent->d_name;
+
+      string filepath = ent->d_name;
+
+      if (log_directory.at(log_directory.size() - 1) == dir_delim_) {
```
sergiud, line 1373: `at` throws on out of range access. This should be avoided.

src/logging.cc
```diff
@@ -1435,19 +1439,19 @@ bool LogCleaner::IsLogFromCurrentProject(const string& filepath,
   for (size_t i = cleaned_base_filename.size(); i < real_filepath_size; i++) {
     const char& c = filepath[i];
 
-    if (i <= cleaned_base_filename.size() + 7) { // 0 ~ 7 : YYYYMMDD
+    if (i <= cleaned_base_filename.size() + 7) {  // 0 ~ 7 : YYYYMMDD
```
sergiud, line 1442: These formatting changes seem to be unrelated. I would undo them.

sergiud (changes requested):
I believe there should be no difference between
```cpp
google::SetLogDestination(google::INFO, "./");
```
and
```cpp
google::SetLogDestination(google::INFO, ".");
```
since `.` and `./` are both directories.

sergiud:
Any progress?

sergiud (approved):

PR #656, opened 2021-05-23
Fix cmake configuration for cygwin environment

romange:
1. Use CYGWIN variable to test for WIN32 because WIN32 var has been deprecated in later versions of cmake.
2. Translate TEST_SRC_DIR to windows paths because posix open() calls can not handle unix-like paths in cygwin environment.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -466,7 +466,16 @@ else (WITH_THREADS AND Threads_FOUND)
   set (NO_THREADS 1)
 endif (WITH_THREADS AND Threads_FOUND)
 
-set (TEST_SRC_DIR \"${CMAKE_CURRENT_SOURCE_DIR}\")
+# fopen/open on Cygwin can not handle unix-type paths like /home/.... 
+# therefore we translate TEST_SRC_DIR to windows-path.
+if (CYGWIN)
+  execute_process (COMMAND cygpath.exe -m ${CMAKE_CURRENT_SOURCE_DIR}
+                   OUTPUT_STRIP_TRAILING_WHITESPACE
+                   OUTPUT_VARIABLE TEST_SRC_DIR)
+  set (TEST_SRC_DIR \"${TEST_SRC_DIR}\")
+else (CYGWIN)
+  set (TEST_SRC_DIR \"${CMAKE_CURRENT_SOURCE_DIR}\")
+endif (CYGWIN)
 
 configure_file (src/config.h.cmake.in config.h)
 configure_file (src/glog/logging.h.in glog/logging.h @ONLY)
@@ -519,16 +528,16 @@ set (GLOG_SRCS
   src/vlog_is_on.cc
 )
 
-if (HAVE_PTHREAD OR WIN32)
+if (HAVE_PTHREAD OR WIN32 OR CYGWIN)
   list (APPEND GLOG_SRCS src/signalhandler.cc)
-endif (HAVE_PTHREAD OR WIN32)
+endif (HAVE_PTHREAD OR WIN32 OR CYGWIN)
 
-if (WIN32)
+if (CYGWIN OR WIN32)
   list (APPEND GLOG_SRCS
     src/windows/port.cc
     src/windows/port.h
   )
-endif (WIN32)
+endif (CYGWIN OR WIN32)
 
 add_compile_options ($<$<AND:$<BOOL:${HAVE_NO_UNNAMED_TYPE_TEMPLATE_ARGS}>,$<NOT:$<CXX_COMPILER_ID:GNU>>>:-Wno-unnamed-type-template-args>)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -466,7 +466,16 @@ else (WITH_THREADS AND Threads_FOUND)
   set (NO_THREADS 1)
 endif (WITH_THREADS AND Threads_FOUND)
 
-set (TEST_SRC_DIR \"${CMAKE_CURRENT_SOURCE_DIR}\")
+# fopen/open on Cygwin can not handle unix-type paths like /home/.... 
+# therefore we translate TEST_SRC_DIR to windows-path.
+if (CYGWIN)
+  EXECUTE_PROCESS(COMMAND cygpath.exe -m ${CMAKE_CURRENT_SOURCE_DIR}
```
sergiud, line 472: I'm not sure this is really necessary. Is it possible that this happens if one uses a non-Cygwin distributed CMake for Cygwin builds?

Besides that, glog conventions for CMake formatting should be used (lowercase, space between command name and parenthesis).

CMakeLists.txt
```diff
@@ -466,7 +466,16 @@ else (WITH_THREADS AND Threads_FOUND)
...
+                  OUTPUT_VARIABLE TEST_SRC_DIR)
+  string (STRIP ${TEST_SRC_DIR} TEST_SRC_DIR)
```
sergiud, line 474: Can be replaced by the `OUTPUT_STRIP_TRAILING_WHITESPACE` property in `execute_process`.

CMakeLists.txt
```diff
@@ -466,7 +466,16 @@ else (WITH_THREADS AND Threads_FOUND)
...
+  set (TEST_SRC_DIR \"${TEST_SRC_DIR}\")
```
sergiud, line 475: This should not be necessary.

CMakeLists.txt
```diff
@@ -466,7 +466,16 @@ else (WITH_THREADS AND Threads_FOUND)
...
+else()
```
sergiud, line 476: Leave a space, add condition.

CMakeLists.txt
sergiud, line 472: Shouldn't `get_filename_component` do the job?

sergiud (changes requested):

CMakeLists.txt
romange: I installed cmake via cygwin package manager. Without this fix TEST_SRC_DIR is defined in config.h as `"/home/roman/glog"`. 
This, however, fails at googletest.h:446 because fopen() can not handle such path style.
With this fix TEST_SRC_DIR  is exported as `"C:/cygwin64/home/roman/glog"`

CMakeLists.txt
romange: it is actually. we need to add double brackets similarly to original `set (TEST_SRC_DIR \"${CMAKE_CURRENT_SOURCE_DIR}\")` line. We need `#define TEST_SRC_DIR "C:/cygwin64/home/roman/glog"`, not `#define TEST_SRC_DIR C:/cygwin64/home/roman/glog`

CMakeLists.txt
romange: I have not find a way to translate it with cmake native functions.

sergiud: You mean the command produces the same incorrect path?

CMakeLists.txt
romange: done

CMakeLists.txt
romange: done

romange:
PTAL

CMakeLists.txt
romange: it's correct in linux terms. cmake is not aware that "/home/..." path is not good for passing into API calls on cygwin. cygwin shell translates the path before calling posix emulation layer. However if you call posix API directly in cygwin, you must use "c:/...." paths. 

Here is minimal example that demonstrates my point:
```
#include <stdio.h>
#include <sys/stat.h>
#include <fcntl.h>

int main() {
    int fd = open("/home/roman/glog/src/config.h.cmake.in", 0, O_RDONLY);
    if (fd < 0) {
        perror("could not open");
    }
    return 0;
}
```

romange: this fails. if you prefix it with "c:/cygwin64/" it passes. 
cmake get_filename_component  just breaks a path into components, it does not produce windows compliant paths.

sergiud: You can obtain at the absolute path using `get_filename_component(... ABSOLUTE)`. There's also `file (TO_CMAKE_PATH ...)` and to `file (TO_NATIVE_PATH ...)`.

I would rather use the native commands if possible. We can leave it as is if other options do not work.

romange: I've checked all these options as well as `REALPATH` before choosing execute_process. Unfortunately, none of them translate to windows native paths.

sergiud (approved):

PR #696, opened 2021-08-08
Add an static cast to compare unsigned with unsigned

Fettpet:
Problem:
The count variable is an unsigned integer. The result of the numeric_limit::max() call is a signed integer. The gcc with -Wall parameter warns about this compare.

Solution:
The result of the numeric_limit::max call is cast to an unsigned value. This prevents that the gcc warns about the compare.


Fettpet:
@googlebot I signed it!

Fettpet:
@googlebot I fixed it.

sergiud:
There's a typo in your commit message: "a~n~ static [...]".

sergiud:
Thanks!

PR #699, opened 2021-08-12
Fix: not implement virtual class when WITH_CUSTOM_PREFIX on

Starsss:


Starsss:
@googlebot I fixed it.

Starsss:
@googlebot I fixed it.

Issue #629, opened 2021-03-31
LLVM libunwind not detected


Arfrever:
There are at least 2 libunwind implementations:
* Original libunwind (https://www.nongnu.org/libunwind/, https://savannah.nongnu.org/projects/libunwind)
* LLVM libunwind (https://github.com/llvm/llvm-project/tree/main/libunwind)

LLVM libunwind provides `libunwind.so`, but it does not have separate platform-specific library (e.g. `libunwind-x86_64.so`).

glog's `cmake/FindUnwind.cmake` assumes that when platform-specific library is not found, then generally libunwind is not found:
```
...
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- Could NOT find Unwind (missing: Unwind_PLATFORM_LIBRARY) 
-- Looking for C++ include dlfcn.h
-- Looking for C++ include dlfcn.h - found
-- Looking for C++ include execinfo.h
-- Looking for C++ include execinfo.h - found
...
```

Workaround working for me is to pass `-DUnwind_PLATFORM_LIBRARY=/usr/lib64/libunwind.so` to cmake.

sergiud:
Since LLVM libunwind already uses CMake, they [should provide](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#find-module-packages) a CMake package config as well. To some extent, this is applies to the original libunwind too. However, CMake support for the latter was added [only recently](https://github.com/libunwind/libunwind/commit/bac2b10878f729f9eaa9076f879bf40f1d58174b#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a).

I'm unwilling to implement workarounds based on heuristics that can easily break. Particularly the LLVM project is very well-known for regularly introducing breaking changes.

I'll mark the issue as won't fix until there are further developments.

Arfrever:
With original libunwind, when I build glog (current trunk) with disabled usage of libunwind platform library (`libunwind-x86_64.so.8` here) (using the same `-DUnwind_PLATFORM_LIBRARY=/usr/lib64/libunwind.so` workaround), then glog builds successfully and its test suite passes.

Original libunwind provides several libraries, including `libunwind-generic.a` and `libunwind-generic.so` symbolic links to `libunwind-${platform}.a` and `libunwind-${platform}.so`:
[...]

PR #708, opened 2021-09-04
Support stack unwind on Android

huangqinjin:


sergiud:
Thanks for the contribution. However, many of the changes are more than questionable. Therefore, please provide a description of the purpose of this PR.

CMakeLists.txt
```diff
@@ -89,7 +89,8 @@ find_package (Unwind)
 
 if (Unwind_FOUND)
   set (HAVE_LIB_UNWIND 1)
-  set (HAVE_UNWIND_H 1)
+else (Unwind_FOUND)
+  check_include_file_cxx (unwind.h HAVE_UNWIND_H)
```
sergiud, line 93: Why do you need this? `FindUnwind.cmake` already accomplishes this.

bazel/glog.bzl
```diff
@@ -128,7 +128,7 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
             "src/stacktrace_powerpc-inl.h",
             "src/stacktrace_windows-inl.h",
             "src/stacktrace_x86-inl.h",
-            "src/stacktrace_x86_64-inl.h",
```
sergiud, line 131: Why did you remove the header from the list?

cmake/FindUnwind.cmake
```diff
@@ -6,8 +6,8 @@
 
 include (FindPackageHandleStandardArgs)
 
-find_path (Unwind_INCLUDE_DIR NAMES unwind.h libunwind.h DOC "unwind include directory")
-find_library (Unwind_LIBRARY NAMES unwind DOC "unwind library")
+find_path (Unwind_INCLUDE_DIR NAMES libunwind.h DOC "libunwind include directory")
```
sergiud, line 9: This is a breaking change that cannot be accepted.

src/logging.cc
```diff
@@ -779,7 +779,8 @@ static void WriteToStderr(const char* message, size_t len) {
 }
 
 inline void LogDestination::MaybeLogToStderr(LogSeverity severity,
-					     const char* message, size_t message_len, size_t /*prefix_len*/) {
+					     const char* message, size_t message_len, size_t prefix_len) {
```
sergiud, line 782: The parameter name is uncommented but still unused. What's the purpose of this change then?

src/logging.cc
```diff
@@ -1868,6 +1846,11 @@ void LogMessage::SendToLog() EXCLUSIVE_LOCKS_REQUIRED(log_mutex) {
     if (write(STDERR_FILENO, message, strlen(message)) < 0) {
       // Ignore errors.
     }
+#if defined(__ANDROID__)
+    __android_log_write(ANDROID_LOG_FATAL,
```
sergiud, line 1850: It is not clear why messages written to logcat are always of fatal severity.

src/utilities.cc
```diff
@@ -92,6 +95,11 @@ static void DebugWriteToStderr(const char* data, void *) {
   if (write(STDERR_FILENO, data, strlen(data)) < 0) {
     // Ignore errors.
   }
+#if defined(__ANDROID__)
+  __android_log_write(ANDROID_LOG_FATAL,
```
sergiud, line 99: As before: it is not clear why messages written to logcat are always of fatal severity.

sergiud (changes requested):

CMakeLists.txt
huangqinjin: `FindUnwind.cmake` is for libunwind library. It has `libunwind.h` and `libunwind.so`. And its API is `unw_*` and is low level. `unwind.h` contains high level API `_Unwind_*` (based on `unw_*` in case of [LLVM](https://github.com/llvm/llvm-project/tree/main/libunwind/include)).

bazel/glog.bzl
huangqinjin: I renamed it to `stacktrace_unwind-inl.h` as it actually uses `_Unwind_*` API rather than hand-coded like `stacktrace_x86-inl.h`.

CMakeLists.txt
sergiud: Please check #629 for a related discussion before making further adjustments.

src/logging.cc
huangqinjin: For Android. https://github.com/google/glog/blob/9733727be4aed835e21a8f73c403f546b71663fe/src/logging.cc#L798

src/utilities.cc
huangqinjin: Because here is fatal error and is going to print stack trace.

src/logging.cc
huangqinjin: Because here is fatal error and is going to print stack trace.

src/logging.cc
sergiud: I see.  Could you please then move the `void` expression to the `#if` block that does not use the parameter?

src/logging.cc
sergiud: Could please add this as comment above then?

CMakeLists.txt
huangqinjin: `HAVE_LIB_UNWIND` includes `libunwind.h` and `HAVE_UNWIND_H` includes `unwind.h`. They are irrelevant in current code:

https://github.com/google/glog/blob/9733727be4aed835e21a8f73c403f546b71663fe/src/utilities.h#L108-L114

cmake/FindUnwind.cmake
huangqinjin: Here we search for `libunwind.h` and `libunwind.so` and will define `HAVE_LIB_UNWIND` if found and use `unw_*` API in code. `unwind.h` does not declare `unw_*` but `_Unwind_*`.

CMakeLists.txt
huangqinjin: Though Android NDK uses LLVM, only `unwind.h` is shipped.

cmake/FindUnwind.cmake
sergiud: This may be true. However, `FindUnwind.cmake` is intended to locate `unwind.h` not `libunwind.h`. The former is included by `stacktrace_x86_64-inl.h`:
https://github.com/google/glog/blob/9733727be4aed835e21a8f73c403f546b71663fe/src/stacktrace_x86_64-inl.h#L35

This is why I argue this change is not acceptable.

huangqinjin: > FindUnwind.cmake is intended to locate unwind.h not libunwind.h

It is not the current logic.

https://github.com/google/glog/blob/9733727be4aed835e21a8f73c403f546b71663fe/CMakeLists.txt#L88-L93


And  `HAVE_LIB_UNWIND` takes precedence over `HAVE_UNWIND_H`. So `unwind.h` is not used if `Unwind_FOUND`.

https://github.com/google/glog/blob/9733727be4aed835e21a8f73c403f546b71663fe/src/utilities.h#L108-L114

sergiud: Okay, but how are these changes (and renaming of headers) are related to Android support?

huangqinjin: only `unwind.h` is shipped with Android NDK.

huangqinjin: I renamed `stacktrace_x86_64-inl.h` to `stacktrace_unwind-inl.h` as it actually uses `_Unwind_*` API rather than hand-coded like `stacktrace_x86-inl.h`.

huangqinjin: Changes to `FindUnwind.cmake` actually are not related to Android support. Just correct it.

sergiud: Got it. I prefer moving `FindUnwind.cmake` changes to a separate PR.

huangqinjin: ok. what about the renaming change?

sergiud: I would revert that too for now. But I agree that we should probably rename the header to make its purpose clearer. Maybe in the same PR as changes to `FindUnwind.cmake`.

sergiud:
Is there a chance you could setup a Github action that uses (any) Android NDK? Without a CI pipeline, it is very likely that the Android support will break again.

huangqinjin:
OK, I will make a separate PR for Android CI.

sergiud:
Thank you. A separate PR is not needed though. You can add the Github action as part of this PR. This will also ensure your changes actually work.

huangqinjin:
CI added. I disabled tests since running tests requires emulators.

.github/workflows/android-build.yml
```diff
@@ -0,0 +1,41 @@
...
+        cmake -S . -B ${{runner.workspace}}/build_CXX${{matrix.std}}-${{matrix.abi}} \
+              -G "Ninja Multi-Config" \
+              -DCMAKE_CXX_EXTENSIONS=OFF \
+              -DCMAKE_CXX_STANDARD=${{matrix.std}} \
+              -DCMAKE_CXX_STANDARD_REQUIRED=ON \
+              -DCMAKE_TOOLCHAIN_FILE=${ANDROID_NDK_HOME}/build/cmake/android.toolchain.cmake \
+              -DANDROID_STL=c++_shared \
+              -DANDROID_NATIVE_API_LEVEL=24 \
+              -DANDROID_ABI=${{matrix.abi}} \
+              -DBUILD_TESTING=OFF
```
sergiud, line 36: I would still build the tests to ensure at least linking works.

.github/workflows/android-build.yml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  build:
+    name: NDK-C++${{matrix.std}}-${{matrix.abi}}-${{matrix.build_type}}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+          std: [98, 11, 14, 17, 20]
+          abi: [arm64-v8a, armeabi-v7a, x86_64, x86]
+          build_type: [Debug, Release, RelWithDebInfo]
```
sergiud, line 14: Is `RelWithDebInfo` actually necessary? I'd say `Debug` and maybe `Release` are sufficient.

CMakeLists.txt
sergiud, line 93: I still do not understand the change. `FindUnwind.cmake` can already tell if `unwind.h` is present.

sergiud:
Thanks!

sergiud (approved):

PR #716, opened 2021-10-01
export `OS_*` defines (fixes #715)

### LINKED ISSUE #715: OS_* defines not always set correctly
`logging.h.in` depends on `OS_*` defines: https://github.com/google/glog/blob/8f9ccfe770add9e4c64e9b25c102658e3c763b73/src/glog/logging.h.in#L104. These defines are set in `utilities.h`: https://github.com/google/glog/blob/2412db315841006e5a4856e64bbb3a274dede0d5/src/utilities.h#L37

`utilities.h` includes `logging.h` but not the other way round. Thus, if someone includes `logging.h` without having included `utilities.h` first, these defines are not set, and compilation will fail on e.g. Windows because it tries to compile gcc-specific code.

A fix could be to move out the `OS_*` defines into a separate header file which could get included by both `logging.h.in` and `utilities.h`.

sergiud:
Moved `OS_*` defines to a separate public header and added a `GLOG_` prefix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,12 +1,3 @@
-/*.cmake
-/*.filters
-/*.sln
-/*.vcxproj
-autom4te.cache
+*.orig
+/build*/
 bazel-*
-CMakeCache.txt
-CMakeFiles/
-config.h
-glog-*.tar.gz
-packages/debian-*
-packages/rpm-unknown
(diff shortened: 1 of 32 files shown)


drigz (approved):
Thanks!

draveness:
@sergiud can we release a new version to fix this?

sergiud:
Please see the corresponding milestone. The date can change though.

Issue #717, opened 2021-10-07
`syscall` warning in Bazel build


ArthurBandaryk:
Hey guys! JFYI when I'm trying to build `glog` with `bazel build :glog` in macOS 11.4 I got the following warning:
![Screen Shot 2021-10-07 at 2 27 25 PM](https://user-images.githubusercontent.com/68085133/136375482-6fb331f6-2a20-4192-b2c3-451beba1e200.png)
Is there a way to fix this warning?
ty:)

PR #718, opened 2021-10-07
Fix `syscall` warning in Bazel Build

ArthurBandaryk:
This commit resolves [#717](https://github.com/google/glog/issues/717) issue.

### DIFF
--- a/bazel/glog.bzl
+++ b/bazel/glog.bzl
@@ -26,10 +26,10 @@ expand_template = rule(
 )
 
 def dict_union(x, y):
-  z = {}
-  z.update(x)
-  z.update(y)
-  return z
+    z = {}
+    z.update(x)
+    z.update(y)
+    return z
 
 def glog_library(namespace = "google", with_gflags = 1, **kwargs):
     if native.repository_name() != "@":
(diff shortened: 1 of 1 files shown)


bazel/glog.bzl
```diff
@@ -92,6 +92,8 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
     darwin_only_copts = [
         # For stacktrace.
         "-DHAVE_DLADDR",
+        "-DHAVE_PTHREAD_THREADID_NP",
```
drigz, line 95: Could you add a comment `# Avoid deprecated syscall()`? Otherwise someone might think this is needed for a stacktrace.

drigz (changes requested):
Thanks for the PR!

bazel/glog.bzl
```diff
@@ -92,6 +92,9 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
     darwin_only_copts = [
         # For stacktrace.
         "-DHAVE_DLADDR",
+        # Avoid deprecated syscall().
+        "-DHAVE_PTHREAD_THREADID_NP",
+        "-Werror",
```
sergiud, line 97: I'm not sure it's a good idea to add the flag here and for macOS only. @drigz Is there a way to enable `-Werror` exclusively for CI builds?

drigz:
@ArthurBandaryk I'm not seeing a CLA signed for your GitHub username, is it possible you signed with a different username?

drigz (approved):

PR #721, opened 2021-10-07
Move win `#define sleep` out of `#ifdef _MSC_VER`

rubensf:
sleep (with lowercase 's') is also not defined on MinGW GCC,
so we have to define it even when not using MSVC.

### DIFF
--- a/src/windows/port.h
+++ b/src/windows/port.h
@@ -114,9 +114,6 @@ enum { STDIN_FILENO = 0, STDOUT_FILENO = 1, STDERR_FILENO = 2 };
 #define hash  hash_compare
 #endif
 
-/* Sleep is in ms, on windows */
-#define sleep(secs)  Sleep((secs) * 1000)
-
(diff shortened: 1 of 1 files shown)


rubensf:
Ah, this was my bad.

I'm using glog @ `96a2f23dca4cc7180821ca5f32e526314395d26a` and with `gn` (instead of cmake), so there are no generated headers. Instead, the build caught `src/windows/glog/logging.h` which was removed in #625. That one has the [`#include unistd.h`](https://github.com/google/glog/blob/96a2f23dca4cc7180821ca5f32e526314395d26a/src/windows/glog/logging.h#L51)  removed, so even though I tried `-DHAVE_UNISTD_H`, unistd.h wouldn't be included :/.

```
#if 0
# include <unistd.h>
#endif
```

Somehow I can't find the correct `unistd.h` in https://github.com/mirror/mingw-w64, but my code locally does find it :)

Thanks for the comment!

PR #723, opened 2021-10-11
ci: replace generated headers by templates

sergiud:
Currently, coverage analysis runs particularly on generated headers. However, since these are not readily available for Coveralls their source cannot be displayed prohibiting their online inspection.

This PR patches the coverage files by replacing the paths of generated headers by their template sources.

### DIFF
--- a/.github/workflows/linux-builds.yml
+++ b/.github/workflows/linux-builds.yml
@@ -71,6 +71,13 @@ jobs:
           '*/src/mock-log.h' \
           '/usr/*' \
           --output-file coverage.info
+
+        for file in src/glog/*.h.in; do
+          name=$(basename ${file})
+          name_we=${name%.h.in}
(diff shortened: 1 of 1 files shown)


sergiud:
@drigz I got in touch with the Coveralls support team because the PR comments still do not appear. They were very helpful at tracing the problem to a disabled setting in the web interface. Specifically, the PR comments must be enabled at https://coveralls.io/github/google/glog/settings under "Pull Requests Alerts":
Screen Shot 2021-10-11 at 1 29 17 PM
Could you please check the setting and enable it as necessary?

.github/workflows/linux-builds.yml
```diff
@@ -71,6 +71,13 @@ jobs:
           '*/src/mock-log.h' \
           '/usr/*' \
           --output-file coverage.info
+
+        for file in $(ls src/glog/*.h.in); do
```
drigz, line 75: nit: I don't think `ls` is needed here, `for f in *.h; do` should work just fine.

PR #731, opened 2021-10-28
Restore the stream format after writing the log prefix

### LINKED ISSUE #695: Glog logging issue with Eigen3
when i use `glog` with `Eigen3` matrix, i found that every value in `Eigen3` matrix is pad with '0'.

I run some tests, and find that

1. `glog` with `Eigen3` matrix: **NOT** fine
2. `glog` with `Eigen3` matrix and `std::setfill(' ')`: fine
3. `cout` with `Eigen3` matrix works: fine
4. reset stream fill char with `LogMessage::stream().fill(' ')` works: fine
5. `glog` with `double` value and `std::setw(4)`: **NOT** fine
6. `glog` with `double` value, `std::setfill(' ')` and `std::setw(4)`: fine

and i check the `logging.cc`, found `stream().fill('0')` calling here, why `fill('0')`?. 

shoud we change to `stream().fill(' ')` or let user to customize padding character?

https://github.com/google/glog/blob/master/src/logging.cc#L1588

https://github.com/google/glog/blob/a6a166db069520dbbd653c97c2e5b12e08a8bb26/src/logging.cc#L1216

----

here's the test code and env

glog version: 0.3.5-1

Eigen version: 3.3.4-4

cpp file

```cpp
#include <iostream>
#include <iomanip>
#include <Eigen/Core>
#include <glog/logging.h>

int main(int /*argc*/, char** /*argv*/) {
  Eigen::Vector3d a = Eigen::Vector3d::Random();
  Eigen::Matrix3d m = Eigen::Matrix3d::Random();
  LOG(INFO) << a.transpose();
  LOG(INFO) << std::setfill(' ') << a.transpose();
  std::cout << a.transpose() << std::endl;
  LOG(INFO) << std::setw(4) << 1.0;
  LOG(INFO) << std::setfill(' ') << std::setw(4) << 1.0;
  google::LogMessage l(__FILE__, __LINE__, google::INFO);
  l.stream().fill(' ');
  l.stream() << a.transpose();

  return 0;
}
```

cmake file

```CMake
project(test)
cmake_minimum_required(VERSION 3.17)
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_FLAGS "-O2 -march=native -no-pie")
find_package(Eigen3 REQUIRED)

add_definitions(-DEIGEN_NO_DEBUG)
include_directories(${EIGEN3_INCLUDE_DIR})

add_executable(test1 test1.cpp)
target_link_libraries(test1 
    glog)
```

output

```
$ ./test1
WARNING: Logging before InitGoogleLogging() is written to STDERR
I0806 17:56:02.290221   583 test1.cpp:9] 00.680375 -0.211234 00.566198
I0806 17:56:02.290475   583 test1.cpp:10]  0.680375 -0.211234  0.566198
 0.680375 -0.211234  0.566198
I0806 17:56:02.290553   583 test1.cpp:12] 0001
I0806 17:56:02.290572   583 test1.cpp:13]    1
I0806 17:56:02.290589   583 test1.cpp:14]  0.680375 -0.211234  0.566198
```

meyerj:
Resolves #695:
> when i use `glog` with `Eigen3` matrix, i found that every value in `Eigen3` matrix is pad with '0'.

We had the same issue in the past, and I applied this patch in a fork in early 2020 already.

I do not remember why exactly I had to update `MungeLine()` in `src/googletest.h`. The new implementation tries to find the prefix at any position of the `line`, not only at the beginning of a word, and preserves whitespace. It may have been because of the log lines with multiple spaces for formatted number output in `logging_unittest.err`, which were munged into one by the previous implementation and hence the comparison failed.

### DIFF
--- a/src/googletest.h
+++ b/src/googletest.h
@@ -421,17 +421,21 @@ static inline bool IsLoggingPrefix(const string& s) {
 //     I20200102 030405 logging_unittest.cc:345] RAW: vlog -1
 //  => IYEARDATE TIME__ logging_unittest.cc:LINE] RAW: vlog -1
 static inline string MungeLine(const string& line) {
-  std::istringstream iss(line);
   string before, logcode_date, time, thread_lineinfo;
-  iss >> logcode_date;
-  while (!IsLoggingPrefix(logcode_date)) {
-    before += " " + logcode_date;
-    if (!(iss >> logcode_date)) {
-      // We cannot find the header of log output.
-      return before;
+  std::size_t begin_of_logging_prefix = 0;
+  for ( ; begin_of_logging_prefix + 9 < line.size(); ++begin_of_logging_prefix) {
+    if (IsLoggingPrefix(line.substr(begin_of_logging_prefix, 9))) {
+      break;
     }
   }
-  if (!before.empty()) before += " ";
+  if (begin_of_logging_prefix > 0) {
+    if (begin_of_logging_prefix + 9 == line.size()) {
+      return line;
+    }
+    before = line.substr(0, begin_of_logging_prefix - 1);
+  }
+  std::istringstream iss(line.substr(begin_of_logging_prefix));
+  iss >> logcode_date;
   iss >> time;
   iss >> thread_lineinfo;
   CHECK(!thread_lineinfo.empty());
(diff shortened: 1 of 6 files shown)


sergiud:
Option 1 sounds reasonable. I would avoid option 3.

meyerj:
> Option 1 sounds reasonable. I would avoid option 3.

Applied in https://github.com/google/glog/pull/731/commits/d002d3d099489238f83fef5dfa88e73d02268d0f.

sergiud:
Thanks!

PR #732, opened 2021-10-29
added log cleaner tests and fixed paths

sergiud:
As a preparation for #654.

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -502,7 +502,6 @@ class LogCleaner {
 
   bool enabled_;
   int overdue_days_;
-  char dir_delim_;  // filepath delimiter ('/' or '\\')
 };
 
 LogCleaner log_cleaner;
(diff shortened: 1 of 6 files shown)


src/logging.cc
```diff
@@ -1323,15 +1323,21 @@ void LogCleaner::Disable() {
...
 
-  if (base_filename_selected) {
-    string dir = base_filename.substr(0, base_filename.find_last_of(dir_delim_) + 1);
-    dirs.push_back(dir);
-  } else {
+  if (!base_filename_selected) {
     dirs = GetLoggingDirectories();
+  } else {
+    size_t pos = base_filename.find_last_of(possible_dir_delim, 0,
+                                            sizeof(possible_dir_delim));
```
aesophor, line 1334: Hello @sergiud, I think the arguments passed to `find_last_of` is incorrect:

* line 1333-1334 would only search `base_filename` for possible_dir_delim starting at index 0 for n characters, where n = `sizeof(possible_dir_delim)`
* I think only the first argument is needed.

sergiud: Thanks. Do you have a specific test case in mind that triggers the problem? Or even better, could you submit directly a PR?

aesophor: Yes I have one!

I called `SetLogDestination(INFO, “testdir/”)` and came across this problem. I’ll submit a PR.

```cpp
#include <glog/logging.h>

int main(int argc, char *argv[]) {
  google::SetLogDestination(google::INFO, "testdir/");
  google::EnableLogCleaner(1);
  LOG(INFO) << "msg";
}
```

PR #746, opened 2021-11-25
Fix log filename format

cmsflash:
The log filename format does not match the example in the line below. One of them must be incorrect. This commit changes the format to match the example.


cmsflash:
@

PR #754, opened 2021-12-07
LogCleaner: Fix relative paths and add a new test

aesophor:
## Overview
Currently, if the user sets the log destination using a relative path (e.g., "testdir/"), then the log cleaner will scan "." for logs instead of "testdir/". This PR fixes the problem and adds a new test.

## Root Cause
* The second parameter `pos` of `string::find_last_of` indicates the position at which the search is to finish.
* We should pass `string::npos` instead of `0` at line 1333.

https://github.com/google/glog/blob/3362cc6c79014170f80e1b8a5c74e7fd61fcb5e4/src/logging.cc#L1333-L1341

## Fix

* Pass `string::npos` instead of `0` at line 1333.
* A new test is added (modified from `src/cleanup_with_prefix_unittest.cc`)
  - In ${TEST_DIR}, create a subdirectory named "test_subdir"
  - After executing the unit test, it expects a single log file in ${TEST_DIR}/test_subdir
  - In ${TEST_DIR}, remove the "test_subdir" directory

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -932,10 +932,15 @@ if (BUILD_TESTING)
 
   target_link_libraries (cleanup_immediately_unittest PRIVATE ${_GLOG_TEST_LIBS})
 
-  add_executable (cleanup_with_prefix_unittest
-    src/cleanup_with_prefix_unittest.cc)
+  add_executable (cleanup_with_absolute_prefix_unittest
+    src/cleanup_with_absolute_prefix_unittest.cc)
 
-  target_link_libraries (cleanup_with_prefix_unittest PRIVATE ${_GLOG_TEST_LIBS})
+  target_link_libraries (cleanup_with_absolute_prefix_unittest PRIVATE ${_GLOG_TEST_LIBS})
+
+  add_executable (cleanup_with_relative_prefix_unittest
+    src/cleanup_with_relative_prefix_unittest.cc)
+
+  target_link_libraries (cleanup_with_relative_prefix_unittest PRIVATE ${_GLOG_TEST_LIBS})
 
   set (CLEANUP_LOG_DIR ${CMAKE_CURRENT_BINARY_DIR}/cleanup_tests)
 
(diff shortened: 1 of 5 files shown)


sergiud:
> [...] the log cleaner will not behave as intended.

What does this mean exactly, i.e., what is the intended behavior?

src/logging.cc
```diff
@@ -1327,8 +1327,7 @@ void LogCleaner::Run(bool base_filename_selected,
   if (!base_filename_selected) {
     dirs = GetLoggingDirectories();
   } else {
-    size_t pos = base_filename.find_last_of(possible_dir_delim, 0,
-                                            sizeof(possible_dir_delim));
+    size_t pos = base_filename.find_last_of(possible_dir_delim);
```
sergiud, line 1330: I do not believe this to be a valid fix because `possible_dir_delim` is not null-terminated. Thus you cannot leave the last `count` parameter out. According to [cppreference](https://en.cppreference.com/w/cpp/string/basic_string/find_last_of), this parameter does not specify how many characters are searched for in the source string but rather in the character string identifying characters to search for.

aesophor: Ah, sorry, I confused the second parameter `pos` of `std::string::find_last_of` with that of `std::string::find_first_of`.

Only the second argument must be changed (from 0 to string::npos)

aesophor:
> > [...] the log cleaner will not behave as intended.
> 
> What does this mean exactly, i.e., what is the intended behavior?

I've updated the description of this PR.

if the user sets the log destination using a relative path (e.g., "testdir/"), then the log cleaner will scan "." for logs instead of "testdir/"

CMakeLists.txt
```diff
@@ -964,6 +976,7 @@ if (BUILD_TESTING)
   # Fixture requirements
   set_tests_properties (cleanup_immediately PROPERTIES FIXTURES_REQUIRED logcleanuptest)
   set_tests_properties (cleanup_with_prefix PROPERTIES FIXTURES_REQUIRED logcleanuptest)
```
sergiud, line 978: It would be adequate now to rename the test and related files to `cleanup_with_absolute_prefix`.

sergiud (changes requested):
Could you please also rebase onto current master and squash all the commits?

sergiud:
Thanks!

PR #760, opened 2021-12-13
LogCleaner: make `overdue_days_` unsigned int

aesophor:
Since the value of `LogCleaner::overdue_days_` should be >= 0,
we can simply make it an unsigned int which also avoids
unnecessary assertions.

Signed-off-by: Marco Wang \<m.aesophor@gmail.com\>


sergiud:
Thanks!

PR #770, opened 2021-12-22
revert to signed int atomics

### LINKED ISSUE #749: LOG_EVERY_N stops printing when using threads
Hi, 
master is currently not working properly when using `LOG_EVERY_N` and `threads`

minimal example to reproduce:
```cpp
#include <gflags/gflags.h>
#include <chrono>
#include <thread>
#include "glog/logging.h"

void f(int id) {
  while (true) {
    std::this_thread::sleep_for(std::chrono::milliseconds(10));
    LOG_EVERY_N(INFO, 10) << "test: " << id;
  }
}

int main(int argc, char* argv[]) {
  gflags::ParseCommandLineFlags(&argc, &argv, true);
  LOG(INFO) << "before threads";

  std::thread t1(f, 0);
  std::thread t2(f, 1);

  t1.join();
  t2.join();
}
```

This will only print for some time before stopping.

The current release does not experience this issue, so it's probably due to a regression.

context:
glog version: `503e3dec8d1fe071376befc62119a837c26612a3`
compiler: `g++ (Ubuntu 8.4.0-1ubuntu1~18.04) 8.4.0`
bazel: `4.2.1`

sergiud:
Fixes #749

### DIFF
--- a/src/glog/logging.h.in
+++ b/src/glog/logging.h.in
@@ -1101,7 +1101,7 @@ namespace google {
 
 #if @ac_cv_cxx11_atomic@ && __cplusplus >= 201103L
 #define SOME_KIND_OF_LOG_EVERY_N(severity, n, what_to_do) \
-  static std::atomic<unsigned> LOG_OCCURRENCES(0), LOG_OCCURRENCES_MOD_N(0); \
+  static std::atomic<int> LOG_OCCURRENCES(0), LOG_OCCURRENCES_MOD_N(0); \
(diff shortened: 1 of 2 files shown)


sergiud:
@drigz LGTM analysis fails since our update to CMake 3.16. Could you please disable it for now?

drigz:
Done.

drigz (approved):

sergiud:
Thank you!

PR #771, opened 2021-12-22
Add the "FLAG_log_year_in_prefix"

msamoila:
Add the "FLAG_log_year_in_prefix" in order to not write the year in the log line prefix

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -144,6 +144,8 @@ GLOG_DEFINE_string(alsologtoemail, "",
                    "in addition to logfiles");
 GLOG_DEFINE_bool(log_prefix, true,
                  "Prepend the log prefix to the start of each log line");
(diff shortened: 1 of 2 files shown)


src/logging.cc
```diff
@@ -144,6 +144,8 @@ GLOG_DEFINE_string(alsologtoemail, "",
                    "in addition to logfiles");
 GLOG_DEFINE_bool(log_prefix, true,
                  "Prepend the log prefix to the start of each log line");
+GLOG_DEFINE_bool(log_year_in_prefix, true,
+                 "The year should be included in the log prefix");
```
sergiud, line 148: Isn't the following wording sufficient then?
```suggestion
                 "Include the year in the log prefix");
```

msamoila: Modified.

PR #775, opened 2021-12-30
stdcxx: eliminate excessive use of std::string::c_str()

anpol:
It could be more clear and slightly more efficient not to convert an
`std::string` into a C-style string whenever possible.


sergiud:
Thanks!

Issue #753, opened 2021-12-07
EnableLogCleaner() interface is not thread safe?


kimi20071025:
I read the functions of `EnableLogCleaner`，`DisableLogCleaner` and `LogCleaner::enabled()`.

I find the private member `enabled_` of  `LogCleaner` is not lock guard. 

Once one thread call  `EnableLogCleaner`，and other thread is logging, It may lead to undefined behaviour?

sergiud:
Could you please point out the exact location where the problem occurs?

kimi20071025:
> Could you please point out the exact location where the problem occurs?

When one thread is logging to file, it will call the function `LogFileObject::Write()`. The function `LogFileObject::Write()` will call `log_cleaner.enabled()`. The `log_cleaner.enabled()` is visit the `enabled_ ` without lock guarded. 

meanwhile other thread is try to change the `enabled_ ` by call function `EnableLogCleaner` or `DisableLogCleaner ` which will  write the `enabled_ ` without lock guarded. 

so may be it lead to undefined behaviour?  I just in doubt of this. May be I am wrong. 

Thanks for response.

sergiud:
Is it possible that you mean a race condition and not undefined behavior? The former can occur because `EnableLogCleaner` and `DisableLogCleaner` are not atomic. However, I believe the race condition is tolerable because at most what can happen are missed cleanups which are made up for once `EnableLogCleaner` is called again.

cc @aesophor

kimi20071025:
> Is it possible that you mean a race condition and not undefined behavior? The former can occur because `EnableLogCleaner` and `DisableLogCleaner` are not atomic. However, I believe the race condition is tolerable because at most what can happen are missed cleanups which are made up for once `EnableLogCleaner` is called again.
> 
> cc @aesophor

Yes. 
One thread is reading `enabled_ ` while another is writing it. May the race condition leads to undefined behavior ? 

It must not leads to undefined behavior because of  `enabled_ ` is `bool` ?

kimi20071025:
> Is it possible that you mean a race condition and not undefined behavior? The former can occur because `EnableLogCleaner` and `DisableLogCleaner` are not atomic. However, I believe the race condition is tolerable because at most what can happen are missed cleanups which are made up for once `EnableLogCleaner` is called again.
> 
> cc @aesophor

Yes. 
One thread is reading `enabled_ ` while another is writing it. May the race condition leads to undefined behavior ? 

[...]

PR #776, opened 2021-12-30
LogCleaner: avoid scanning logs too frequently

fdgkhdkgh:
This patch is for the performance issue : #753. 

Originally, "LogCleaner" scanned all the log directories every time we flush. In the common case, we flush every 30 seconds and then called "LogCleaner::Run". But if we flush every time we do "LogFileObject::Write" like the test code from @kimi20071025, we'll call "LogCleaner::Run" so many times.

In this patch, I maintain an additional timer `next_cleanup_time_ ` ( thank @aesophor  for the advice  ), and avoid scanning the logs too frequently.

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -154,6 +154,10 @@ GLOG_DEFINE_int32(logbuflevel, 0,
                   " ...)");
 GLOG_DEFINE_int32(logbufsecs, 30,
                   "Buffer log messages for at most this many seconds");
+
+GLOG_DEFINE_int32(logcleansecs, 60 * 5, // every 5 minutes
+                  "Clean overdue logs every this many seconds");
+
 GLOG_DEFINE_int32(logemaillevel, 999,
                   "Email log messages logged at this level or higher"
                   " (0 means email all; 3 means email FATAL only;"
@@ -484,9 +488,12 @@ class LogCleaner {
   void Enable(unsigned int overdue_days);
   void Disable();
 
+  // update next_cleanup_time_
+  void UpdateCleanUpTime();
+
   void Run(bool base_filename_selected,
            const string& base_filename,
-           const string& filename_extension) const;
+           const string& filename_extension);
 
   bool enabled() const { return enabled_; }
 
@@ -503,6 +510,7 @@ class LogCleaner {
 
   bool enabled_;
   unsigned int overdue_days_;
+  int64 next_cleanup_time_;         // cycle count at which to clean overdue log
 };
 
 LogCleaner log_cleaner;
@@ -1303,7 +1311,7 @@ void LogFileObject::Write(bool force_flush,
   }
 }
 
-LogCleaner::LogCleaner() : enabled_(false), overdue_days_(7) {}
+LogCleaner::LogCleaner() : enabled_(false), overdue_days_(7), next_cleanup_time_(0) {}
 
 void LogCleaner::Enable(unsigned int overdue_days) {
   enabled_ = true;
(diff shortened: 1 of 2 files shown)


src/logging.cc
```diff
@@ -1314,12 +1322,24 @@ void LogCleaner::Disable() {
...
 void LogCleaner::Run(bool base_filename_selected,
                      const string& base_filename,
-                     const string& filename_extension) const {
+                     const string& filename_extension) {
   assert(enabled_);
   assert(!base_filename_selected || !base_filename.empty());
 
+  // avoid scanning logs too frequently
+  if (CycleClock_Now() < next_cleanup_time_) {
+    return;
```
sergiud, line 1339: This line is not covered by any tests. Could you please either extend an existing test to cover the line or remove it altogether?

src/logging.cc
```diff
@@ -154,6 +154,10 @@ GLOG_DEFINE_int32(logbuflevel, 0,
                   " ...)");
 GLOG_DEFINE_int32(logbufsecs, 30,
                   "Buffer log messages for at most this many seconds");
+
+GLOG_DEFINE_int32(logcleansecs, 300,
+                  "Clean overdue logs every this many seconds");
```
sergiud, line 159: What's the reason for choosing a signed integer to represent the time span? Isn't it always positive?

sergiud (changes requested):
Thanks for the PR. Please take a look at the [contribution guide](https://github.com/google/glog#how-to-contribute) and make sure to follow it. Particularly, you must sign the CLA.

src/logging.cc
```diff
@@ -154,6 +154,10 @@ GLOG_DEFINE_int32(logbuflevel, 0,
                   " ...)");
 GLOG_DEFINE_int32(logbufsecs, 30,
                   "Buffer log messages for at most this many seconds");
+
+GLOG_DEFINE_int32(logcleansecs, 300,
```
sergiud, line 158:
```suggestion
GLOG_DEFINE_int32(logcleansecs, 60 * 5, // every 5 minutes
```

src/logging.cc
```diff
@@ -503,6 +510,7 @@ class LogCleaner {
 
   bool enabled_;
   unsigned int overdue_days_;
+  int64 next_cleanup_time_;         // cycle count at which to clean overdue log
```
sergiud, line 513: As before: why using a signed integer?

src/logging.cc
```diff
@@ -1314,12 +1322,24 @@ void LogCleaner::Disable() {
   enabled_ = false;
 }
 
+void LogCleaner::UpdateCleanUpTime() {
+  const int64 next = (FLAGS_logcleansecs
+                      * static_cast<int64>(1000000));  // in usec
```
sergiud, line 1327: The literal can be represented as `int` and will be automatically promoted to a 64-bit within the expression. The cast is therefore not necessary.

src/logging.cc
fdgkhdkgh: Thanks for the reminder, I've extended an existing test in file : cleanup_with_absolute_prefix_unittest.cc

src/logging.cc
fdgkhdkgh: Yes, it's always positive.
I use integer because another timer ( e.g. `next_flush_time_` ) use integer, too.

```cpp=
void LogFileObject::FlushUnlocked(){
  if (file_ != NULL) {
    fflush(file_);
    bytes_since_flush_ = 0;
  }
  // Figure out when we are due for another flush.
  const int64 next = (FLAGS_logbufsecs
                      * static_cast<int64>(1000000));  // in usec
  next_flush_time_ = CycleClock_Now() + UsecToCycles(next);
}
```

sergiud:
@fdgkhdkgh Some comments are still open. Please let me know if you need some clarification.

src/logging.cc
fdgkhdkgh: Thanks for the suggestion, I'll fix it.

src/logging.cc
fdgkhdkgh: As before: I use integer because another timer ( e.g. `next_flush_time_` ) use integer, too.

src/logging.cc
fdgkhdkgh: You're right, I'll remove the cast.

aesophor:
@fdgkhdkgh Thank you for adding this!

sergiud:
This looks good to me now (unless @aesophor has any objections).

@fdgkhdkgh Would you like to add yourself to the `AUTHORS` and `CONTRIBUTORS` file as per contribution guide before we merge this PR?

aesophor:
LGTM. Thanks so much! :smile:

sergiud:
Thanks!

sergiud (approved):

Issue #57, opened 2015-09-07
libglog.pc.in should include libunwind as private dep for static linking case


Katerall:
This was reported along with a patch by someone before (+CLA):
https://code.google.com/p/google-glog/issues/detail?id=207

I ran into this issue during compilation with errors similar to those reported in the issue:

```
/local/adp/unwind-compile-error/protobuf-util/vendor/src/glog-build/../glog/src/stacktrace_libunwind-inl.h:65: undefined reference to `_Ux86_64_getcontext'
/local/adp/unwind-compile-error/protobuf-util/vendor/src/glog-build/../glog/src/stacktrace_libunwind-inl.h:66: undefined reference to `_ULx86_64_init_local'
/local/adp/unwind-compile-error/protobuf-util/vendor/src/glog-build/../glog/src/stacktrace_libunwind-inl.h:78: undefined reference to `_ULx86_64_step'
/local/adp/unwind-compile-error/protobuf-util/vendor/src/glog-build/../glog/src/stacktrace_libunwind-inl.h:70: undefined reference to `_ULx86_64_get_reg'
```

Applying the patch worked for me. Please fix.

amosbird:
The patch gets lost in the original url. I don't see related fixes, however, it's mentioned as fixed in 0.5. Any references?

sergiud:
The issue was closed without a fix. Please submit a PR.

PR #787, opened 2022-01-24
libglog.pc: Set Libs.private for static linking

Arfrever:
This implements suggestion from #57.

Before:
```
$ pkg-config --libs libglog
-lglog
$ pkg-config --libs --static libglog
-lglog
```

After:
```
$ pkg-config --libs libglog
-lglog
$ pkg-config --libs --static libglog
-lglog -lunwind -lpthread -lgflags
```

By the way, I was also trying `get_target_property(some_output_variable glog LINK_LIBRARIES)`, but it provides raw cmake-style value such as `unwind::unwind;-lpthread;gflags` which would need complicated parsing and transformation, so the solution used in this pull request seems easier.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -533,26 +533,6 @@ configure_file (src/glog/raw_logging.h.in glog/raw_logging.h @ONLY)
 configure_file (src/glog/stl_logging.h.in glog/stl_logging.h @ONLY)
 configure_file (src/glog/vlog_is_on.h.in glog/vlog_is_on.h @ONLY)
 
-if (WITH_PKGCONFIG)
-  set (VERSION ${PROJECT_VERSION})
-  set (prefix ${CMAKE_INSTALL_PREFIX})
-  set (exec_prefix ${CMAKE_INSTALL_FULL_BINDIR})
-  set (libdir ${CMAKE_INSTALL_FULL_LIBDIR})
-  set (includedir ${CMAKE_INSTALL_FULL_INCLUDEDIR})
-
-  configure_file (
-    "${PROJECT_SOURCE_DIR}/libglog.pc.in"
-    "${PROJECT_BINARY_DIR}/libglog.pc"
-    @ONLY
-  )
-
-  unset (VERSION)
-  unset (prefix)
-  unset (exec_prefix)
-  unset (libdir)
-  unset (includedir)
-endif (WITH_PKGCONFIG)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -714,6 +703,28 @@ generate_export_header (glog
   EXPORT_MACRO_NAME GOOGLE_GLOG_DLL_DECL
   EXPORT_FILE_NAME ${CMAKE_CURRENT_BINARY_DIR}/glog/export.h)
 
+string(STRIP "${glog_libraries_options_for_static_linking}" glog_libraries_options_for_static_linking)
```
sergiud, line 706:
```suggestion
string (STRIP "${glog_libraries_options_for_static_linking}" glog_libraries_options_for_static_linking)
```

CMakeLists.txt
```diff
@@ -641,17 +621,24 @@ add_library (glog
...
 endif (Unwind_FOUND)
 
 if (HAVE_DBGHELP)
   target_link_libraries (glog PRIVATE dbghelp)
+  set (glog_libraries_options_for_static_linking "${glog_libraries_options_for_static_linking} -ldbghelp")
 endif (HAVE_DBGHELP)
 
 if (HAVE_PTHREAD)
   target_link_libraries (glog PRIVATE ${CMAKE_THREAD_LIBS_INIT})
+  if (CMAKE_THREAD_LIBS_INIT)
```
sergiud, line 639: I prefer to leave an empty line before/after conditional blocks for better readability.

```suggestion

  if (CMAKE_THREAD_LIBS_INIT)
```

CMakeLists.txt
```diff
@@ -661,6 +648,7 @@ if (gflags_FOUND)
   else (TARGET gflags::gflags)
     target_link_libraries (glog PUBLIC gflags)
   endif (TARGET gflags::gflags)
+  set (glog_libraries_options_for_static_linking "${glog_libraries_options_for_static_linking} -lgflags")
```
sergiud, line 651:
```suggestion

  set (glog_libraries_options_for_static_linking "${glog_libraries_options_for_static_linking} -lgflags")
```

sergiud (changes requested):
LGTM. Only some nitpicks.

sergiud (approved):

sergiud:
Thanks!

PR #790, opened 2022-02-13
Add the logtostdout and colorlogtostdout flag to allow logging to stdout

### LINKED ISSUE #701: Console logging: separation of stdout and stderr
For console logging, it seems like there is no way to configure Glog to send non-error logs to `stdout` and error logs to `stderr`. 

By default, Glog sends only error logs to stderr, and there is a `FLAGS_logtostderr` which sends ALL logs to `stderr` which is not ideal.  It will be good to have a flag to log to _both_ stdout and stderr appropriately.

git-hulk:
Currently, glog allowed to use logtostderr to send error logs to stderr,
but many log tailors would regard logs from stderr as error logs. So we
want to send non-error logs to stdout and only send error logs to stderr
according to the stderrthreshold.

Resolved #701

### DIFF
--- a/src/googletest.h
+++ b/src/googletest.h
@@ -350,6 +350,9 @@ static inline void CaptureTestOutput(int fd, const string & filename) {
   CHECK(s_captured_streams[fd] == NULL);
   s_captured_streams[fd] = new CapturedStream(fd, filename);
 }
+static inline void CaptureTestStdout() {
+  CaptureTestOutput(STDOUT_FILENO, FLAGS_test_tmpdir + "/captured.out");
+}
 static inline void CaptureTestStderr() {
   CaptureTestOutput(STDERR_FILENO, FLAGS_test_tmpdir + "/captured.err");
 }
@@ -504,9 +507,13 @@ static inline void WriteToFile(const string& body, const string& file) {
   fclose(fp);
 }
 
-static inline bool MungeAndDiffTestStderr(const string& golden_filename) {
-  CapturedStream* cap = s_captured_streams[STDERR_FILENO];
-  CHECK(cap) << ": did you forget CaptureTestStderr()?";
+static inline bool MungeAndDiffTest(const string& golden_filename,
+                                    CapturedStream* cap) {
+  if (cap == s_captured_streams[STDOUT_FILENO]) {
+    CHECK(cap) << ": did you forget CaptureTestStdout()?";
+  } else {
+    CHECK(cap) << ": did you forget CaptureTestStderr()?";
+  }
 
   cap->StopCapture();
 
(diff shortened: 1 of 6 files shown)


sergiud:
The tests of MSVC builds are failing. Could you please rebase onto current master and see if this helps?

git-hulk:
@sergiud cool, thanks for your great help!

git-hulk:
@sergiud How do you think about this PR? Does this implementation make sense for you?

sergiud:
I already skipped through the PR and it does make sense to me. However, I would like to wait for additional comments from users (if any). So please hold tight.

git-hulk:
ok, cool. Sorry for pushing you.

src/googletest.h
```diff
@@ -536,6 +546,14 @@ static inline bool MungeAndDiffTestStderr(const string& golden_filename) {
   return true;
 }
 
+static inline bool MungeAndDiffTestStderr(const string& golden_filename) {
+  return MungeAndDiffTestStderrOrStdout(golden_filename, false);
```
sergiud, line 550: `bool` parameters are hard to follow. It would be clearer if you pass `CapturedStream* cap` or an index into `s_captured_streams` to avoid code repetition.

src/logging.cc
```diff
@@ -739,43 +743,61 @@ inline void LogDestination::SetEmailLogging(LogSeverity min_severity,
   LogDestination::addresses_ = addresses;
 }
 
-static void ColoredWriteToStderr(LogSeverity severity,
-                                 const char* message, size_t len) {
-  const GLogColor color =
-      (LogDestination::terminal_supports_color() && FLAGS_colorlogtostderr) ?
-      SeverityToColor(severity) : COLOR_DEFAULT;
+static void ColoredWriteToStderrOrStdout(bool is_stdout, LogSeverity severity,
```
sergiud, line 746: Instead of a `bool` you can pass a `FILE*` directly.

src/logging.cc
```diff
@@ -739,43 +743,61 @@ inline void LogDestination::SetEmailLogging(LogSeverity min_severity,
...
+                                         const char* message, size_t len) {
+  // We also need to send logs to the stderr when the severity is
+  // higher or equal to the stderr threshold.
+  is_stdout = is_stdout && severity < FLAGS_stderrthreshold;
```
sergiud, line 750: I think the decision whether the output is written `stdout` or `stderr` should be made outside of this function. You can introduce a helper function if you need to.

src/logging.cc
```diff
@@ -739,43 +743,61 @@ inline void LogDestination::SetEmailLogging(LogSeverity min_severity,
...
   // exit code, and cerr may be partially or fully destroyed by then.
   if (COLOR_DEFAULT == color) {
-    fwrite(message, len, 1, stderr);
+    fwrite(message, len, 1, output_fd);
     return;
   }
 #ifdef GLOG_OS_WINDOWS
-  const HANDLE stderr_handle = GetStdHandle(STD_ERROR_HANDLE);
+  const HANDLE output_handle =
+      GetStdHandle(is_stdout ? STD_OUTPUT_HANDLE : STD_ERROR_HANDLE);
```
sergiud, line 767: Here you can distinguish between `stderr` and `stdout` without relying on a `bool`.

sergiud (changes requested):
I have a few suggestions. Thanks for looking into the issue!

sergiud:
Looks great!

sergiud (approved):

Issue #793, opened 2022-02-18
Test suite does not fully respect ${TMPDIR}


Arfrever:
Portage, package manager in Gentoo, sets `TMPDIR` environmental variable to per-package-version temporary directory.
E.g. `TMPDIR="/var/tmp/portage/dev-cpp/glog-0.5.0/temp"`

When test suite of Glog is run, it creates e.g. the following files in `${TMPDIR}`:
```
logging_custom_prefix_unittest.ERROR -> logging_custom_prefix_unittest.localhost.portage.log.ERROR.20220218-105043.71
logging_custom_prefix_unittest.INFO -> logging_custom_prefix_unittest.localhost.portage.log.INFO.20220218-105043.71
logging_custom_prefix_unittest.WARNING -> logging_custom_prefix_unittest.localhost.portage.log.WARNING.20220218-105043.71
logging_custom_prefix_unittest.localhost.portage.log.ERROR.20220218-105043.71
logging_custom_prefix_unittest.localhost.portage.log.INFO.20220218-105043.71
logging_custom_prefix_unittest.localhost.portage.log.WARNING.20220218-105043.71
logging_unittest.ERROR -> logging_unittest.localhost.portage.log.ERROR.20220218-105044.76
logging_unittest.INFO -> logging_unittest.localhost.portage.log.INFO.20220218-105044.76
logging_unittest.WARNING -> logging_unittest.localhost.portage.log.WARNING.20220218-105044.76
logging_unittest.localhost.portage.log.ERROR.20220218-105044.76
logging_unittest.localhost.portage.log.INFO.20220218-105044.76
logging_unittest.localhost.portage.log.WARNING.20220218-105044.76
mock-log_unittest.ERROR -> mock-log_unittest.localhost.portage.log.ERROR.20220218-105046.84
mock-log_unittest.INFO -> mock-log_unittest.localhost.portage.log.INFO.20220218-105046.84
mock-log_unittest.WARNING -> mock-log_unittest.localhost.portage.log.WARNING.20220218-105046.84
mock-log_unittest.localhost.portage.log.ERROR.20220218-105046.84
mock-log_unittest.localhost.portage.log.INFO.20220218-105046.84
mock-log_unittest.localhost.portage.log.WARNING.20220218-105046.84
```

However test suite also leaves some files in `/tmp` without respecting `${TMPDIR}`, e.g.:
```
captured.err
logging_custom_prefix_unittest.INFO -> logging_test_basename_two_processes_writing
[...]

PR #794, opened 2022-02-19
Tests: Respect TEST_TMPDIR, TMPDIR, TMP environmental variables

Arfrever:
Use GetExistingTempDirectories() (which uses GetTempDirectories() which respects
TEST_TMPDIR, TMPDIR, TMP environmental variables) for FLAGS_test_tmpdir variable
used in some tests.

Fixes: #793

### DIFF
--- a/src/googletest.h
+++ b/src/googletest.h
@@ -76,13 +76,16 @@ _END_GOOGLE_NAMESPACE_
 #define GLOG_EXPORT
 
 static inline string GetTempDir() {
-#ifndef GLOG_OS_WINDOWS
-  return "/tmp";
-#else
-  char tmp[MAX_PATH];
-  GetTempPathA(MAX_PATH, tmp);
-  return tmp;
-#endif
+  vector<string> temp_directories_list;
+  google::GetExistingTempDirectories(&temp_directories_list);
+
+  if (temp_directories_list.empty()) {
+    fprintf(stderr, "No temporary directory found\n");
+    exit(EXIT_FAILURE);
+  }
+
+  // Use first directory from list of existing temporary directories.
+  return temp_directories_list.front();
 }
 
 #if defined(GLOG_OS_WINDOWS) && defined(_MSC_VER) && !defined(TEST_SRC_DIR)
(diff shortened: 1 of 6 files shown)


src/googletest.h
```diff
@@ -75,14 +75,21 @@ _END_GOOGLE_NAMESPACE_
...
 static inline string GetTempDir() {
-#ifndef GLOG_OS_WINDOWS
-  return "/tmp";
-#else
-  char tmp[MAX_PATH];
-  GetTempPathA(MAX_PATH, tmp);
-  return tmp;
-#endif
+  if (temp_directories_list == NULL) {
+    temp_directories_list = new vector<string>;
```
sergiud, line 82: I'm not happy with this dynamic memory allocation because it will unnecessarily leak memory. The pointer can be easily changed to a normal `std::vector<std::string>` instance.

src/googletest.h
```diff
@@ -75,14 +75,21 @@ _END_GOOGLE_NAMESPACE_
...
+    google::GetExistingTempDirectories(temp_directories_list);
+
+    if (temp_directories_list->empty()) {
+      fprintf(stderr, "No temporary directory found\n");
+      exit(1);
```
sergiud, line 87:
```suggestion
      exit(EXIT_FAILURE);
```
`EXIT_FAILURE` is defined `<cstdlib>`.

sergiud (changes requested):
Thanks for the PR. Only couple of suggestions.

src/googletest.h
Arfrever: OK. I will also update (in separate commit) other instances in glog source code:
```
$ grep -r 'exit(' *
README.rst:     exit(1);
src/logging_unittest.cc:    exit(0);
src/logging_custom_prefix_unittest.cc:    exit(0);
src/googletest.h:      exit(1);
src/googletest.h:      exit(1);                                                                 \
src/googletest.h:      exit(1);                                          \
src/googletest.h:      exit(1);                                                          \
src/googletest.h:      exit(1);                                                  \
src/googletest.h:      exit(1);                                                  \
src/googletest.h:      exit(1);                                                          \
src/googletest.h:      exit(1);                                                          \
src/googletest.h:      exit(1);                                                          \
src/glog/raw_logging.h.in:    exit(1); \
src/glog/logging.h.in:  @ac_cv___attribute___noreturn@ ~NullStreamFatal() throw () { _exit(1); }
```

sergiud: Thanks!

sergiud (approved):

sergiud:
Thank you!

PR #801, opened 2022-02-25
Fix namespace resolution issue in LOG_EVERY_T

skeptic-monkey:
int64 is a typedef defined within @ac_google_namespace@, while LOG_EVERY_T is defined within ::google:: namespace.
but @ac_google_namespace@ is configurable and might not always be "google".

It can result in a compilation error, int64 being undefined. using absolute namespace resolution with @ac_google_namespace@:: fixes it.

Also replaced 2 other hardcoded google:: usage.


skeptic-monkey:
I stuggle to see the link between my change and the failing tests on MacOS.

sergiud:
Thanks for the PR. The macOS runner seems to be volatile or our tests flaky.

Can you provide a reproducer for the failures you are describing? I'm just wondering why our CI did not catch these problems,

skeptic-monkey:
I honestly don't understand either, I checked and this part of the code seem covered by the tests.

I'll try to create a sample project that reproduce it. Meanwhile I can at least describe our setup.
We are integrating glog in our bazel based project, and we build on windows 11 using clang-cl.

sergiud:
Thanks for the details.

@drigz It seems we need a Bazel runner that uses `clang-cl` in addition to MSVC. Is there a chance you can look into it?

sergiud:
@skeptic-monkey Could you please rebase?

skeptic-monkey:
> @skeptic-monkey Could you please rebase?

Done.

PR #802, opened 2022-03-02
Add clang-cl.exe to Bazel CI

drigz:
This should catch issues like #801. This uses the new `tasks` syntax to
define multiple Windows tasks:

https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md#basic-syntax

I've removed `/WX` to allow warnings because there are quite a few, and I'm not confident enough in my C++ to dig into solving them. https://buildkite.com/bazel/google-logging/builds/2611#8de3f805-297c-484e-a63d-2092e8e49faa

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -51,7 +51,7 @@ tasks:
     - "--extra_toolchains=@local_config_cc//:cc-toolchain-x64_windows-clang-cl"
     - "--extra_execution_platforms=//:x64_windows-clang-cl"
     - "--features=layering_check"
-    - "--copt=/WX"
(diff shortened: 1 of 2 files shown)


sergiud:
Thanks. Looks great! Only some runners are having a bad day.

sergiud (approved):

sergiud:
@drigz You can try amending the latest commit and force push the changes. This should solve Codecov issues.

sergiud:
I think we can ignore thore macOS runner failure. I'll look into it later.

drigz:
Thanks for checking. Do you also plan to look at the clang warnings?

sergiud:
Yes, I'll look at that too. Thanks!

PR #803, opened 2022-03-07
Fix "'GLOG_EXPORT' macro redefined" on clang-cl

drigz:
The previous approach used
--incompatible_enable_cc_toolchain_resolution, which is recommended by
the docs, but a Bazel developer told me it's obsolete. The new, old
approach is simpler and should stop the warning from being user-visible.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -47,18 +47,12 @@ tasks:
     environment:
       BAZEL_VC: "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\VC"
     build_flags:
-    - "--incompatible_enable_cc_toolchain_resolution"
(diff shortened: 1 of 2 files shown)


sergiud:
I wondering if it is possible to split `GLOG_EXPORT` definitions using [`local_defines`](https://docs.bazel.build/versions/1.2.0/be/c-cpp.html#cc_binary.local_defines) to avoid defining `-Wno-macro-redefined`?

drigz:
Bazel doesn't appear to be smart enough to deduplicate `defines` and `local_defines`:

```
        local_defines = ["FOO=local"],
        defines = ["FOO=global"],
```

build with `-s` leads to:

```
/usr/bin/gcc [...] '-DFOO=global' [...] '-DFOO=local' [...]
```

sergiud (approved):

Issue #809, opened 2022-03-21
Support rolling log files base on file size and time 


yangzhg:
Hi, it seems that glog currently only supports scrolling log files by file size, not by time at the same time, in many scenarios there is a need to support slicing log files by time.

Also, glog can only clean up history log files by time, not by number of history log files. 

So I would like to submit a pr that supports splitting log files by time and cleaning history files by number of files.

PR #811, opened 2022-03-22
Support rolling log files base on file size and time

yangzhg:
This pr implements #809
Add two flags max_logfile_num and log_rolling_policy, support rolling
log file by day or by hour, and use max_logfile_num to control the max
history log file num after rolling.

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -58,11 +58,12 @@
 #ifdef HAVE_SYSLOG_H
 # include <syslog.h>
 #endif
-#include <vector>
-#include <cerrno>                   // for errno
-#include <sstream>
+#include <cctype>  // for std::isspace
+#include <cerrno>  // for errno
 #include <regex>
-#include <cctype> // for std::isspace
+#include <set>
+#include <sstream>
+#include <vector>
 #ifdef GLOG_OS_WINDOWS
 #include "windows/dirent.h"
 #else
@@ -81,15 +82,16 @@
 #include <android/log.h>
 #endif
 
-using std::string;
-using std::vector;
-using std::setw;
-using std::setfill;
-using std::hex;
 using std::dec;
+using std::hex;
 using std::min;
+using std::multiset;
 using std::ostream;
 using std::ostringstream;
+using std::setfill;
+using std::setw;
+using std::string;
+using std::vector;
 
 using std::FILE;
 using std::fwrite;
@@ -200,6 +202,15 @@ GLOG_DEFINE_string(log_link, "", "Put additional links to the log "
 GLOG_DEFINE_uint32(max_log_size, 1800,
                    "approx. maximum log file size (in MB). A value of 0 will "
                    "be silently overridden to 1.");
+
+GLOG_DEFINE_string(
+    log_rolling_policy, "size",
+    "log file rolling policy, support size-based and time-based."
+    " The available values are size, day and hour, if use time-based policy,"
+    " the maximum log file size also control by max_log_size");
+
+GLOG_DEFINE_uint32(max_logfile_num, 10,
+                   "maximum history log file num after rolling");
 
 GLOG_DEFINE_bool(stop_logging_if_full_disk, false,
                  "Stop attempting to log to disk if the disk is full.");
@@ -433,6 +444,12 @@ void* custom_prefix_callback_data = nullptr;
 
 namespace {
 
+struct Filetime {
+  std::string name;
+  std::time_t time;
+  bool operator<(const Filetime& o) const { return time < o.time; }
+};
+
 // Encapsulates all file-system related state
 class LogFileObject : public base::Logger {
  public:
@@ -463,6 +480,10 @@ class LogFileObject : public base::Logger {
   // acquiring lock_.
   void FlushUnlocked();
 
+  // check the logfiles rolling by size or date, and remve the oldest ones
+  // according to FLAGS_max_logfile_num
+  void CheckHistoryFileNum();
+
  private:
   static const uint32 kRolloverAttemptFrequency = 0x20;
 
@@ -476,14 +497,24 @@ class LogFileObject : public base::Logger {
   uint32 bytes_since_flush_{0};
   uint32 dropped_mem_length_{0};
   uint32 file_length_{0};
-  unsigned int rollover_attempt_;
+  uint32 rollover_attempt_;
   int64 next_flush_time_{0};  // cycle count at which to flush log
   WallTime start_time_;
+  std::multiset<Filetime> file_list_;
+  bool initialized_;
+  std::tm tm_time_;
 
   // Actually create a logfile using the value of base_filename_ and the
   // optional argument time_pid_string
   // REQUIRES: lock_ is held
   bool CreateLogfile(const string& time_pid_string);
+
+  // Actually create a logfile using full file name and flags
+  bool CreateLogfileInternal(const char* filename, int flags);
+
+  // Check if log file is getting too big or match the date time condition, If
+  // so, rollover.
+  bool CheckNeedRollLogFiles(time_t timestamp);
 };
 
 // Encapsulate all log cleaner related states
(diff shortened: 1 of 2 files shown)


src/logging.cc
```diff
@@ -428,6 +438,13 @@ namespace  {
 #endif
 
 namespace {
+typedef struct filetime {
+  std::string name;
+  time_t time;
+  bool operator<(const struct filetime& o) const {
+    return o.time > time;
+  }
+} Filetime;
```
sergiud, line 447:
```suggestion
struct Filetime {
  std::string name;
  std::time_t time;
  bool operator<(const struct Filetime& o) const {
    return time < o.time;
  }
};
```
It's also not clear what `time` refers to. Is it the creation, access, or modification time? Is `name` the full path? Brief comments above each member are sufficient.

src/logging.cc
```diff
@@ -1138,6 +1163,54 @@ bool LogFileObject::CreateLogfile(const string& time_pid_string) {
   return true;  // Everything worked
 }
 
+void LogFileObject::CheckHistoryFileNum() {
+  struct dirent *entry;
+  DIR *dp;
+
+  const vector<string> &log_dirs = GetLoggingDirectories();
+  if (log_dirs.size() < 1) return;
```
sergiud, line 1171:
```suggestion
  if (log_dirs.empty()) return;
```

src/logging.cc
```diff
@@ -476,6 +494,9 @@ class LogFileObject : public base::Logger {
   unsigned int rollover_attempt_;
   int64 next_flush_time_;         // cycle count at which to flush log
   WallTime start_time_;
+  std::list<Filetime> file_list_;
```
sergiud, line 497: What's the reason for choosing a `std::list` here? Looks like you want a `std::priority_queue` here since you are constantly sorting the container elements.

Also the member variable name should indicate what kind of times are stored.

src/logging.cc
```diff
@@ -476,6 +494,9 @@ class LogFileObject : public base::Logger {
...
+  bool inited_;
```
sergiud, line 498:
```suggestion
  bool initialized_;
```

src/logging.cc
```diff
@@ -1138,6 +1163,54 @@ bool LogFileObject::CreateLogfile(const string& time_pid_string) {
...
+    return;
+  }
+
+  file_list_.clear();
+  while ((entry = readdir(dp)) != NULL) {
+    if (DT_DIR == entry->d_type ||
+        DT_LNK == entry->d_type) {
+      continue;
+    }
+    std::string filename = std::string(entry->d_name);
```
sergiud, line 1186:
```suggestion
    std::string filename = entry->d_name;
```

src/logging.cc
```diff
@@ -459,6 +476,7 @@ class LogFileObject : public base::Logger {
   // can avoid grabbing a lock.  Usually Flush() calls it after
   // acquiring lock_.
   void FlushUnlocked();
+  void CheckHistoryFileNum();
```
sergiud, line 479: A brief comment describing the purpose of the method would be helpful here.

src/logging.cc
```diff
@@ -1148,13 +1221,50 @@ void LogFileObject::Write(bool force_flush,
   if (base_filename_selected_ && base_filename_.empty()) {
     return;
   }
-
-  if (file_length_ >> 20U >= MaxLogSize() || PidHasChanged()) {
+  struct ::tm tm_time;
+  bool is_split = false;
+  if ("day" == FLAGS_log_rolling_policy) {
```
sergiud, line 1226:
```suggestion
  if (FLAGS_log_rolling_policy == "day") {
```
and below too.

src/logging.cc
```diff
@@ -1148,13 +1221,50 @@ void LogFileObject::Write(bool force_flush,
   if (base_filename_selected_ && base_filename_.empty()) {
     return;
   }
-
-  if (file_length_ >> 20U >= MaxLogSize() || PidHasChanged()) {
+  struct ::tm tm_time;
```
sergiud, line 1224: The following block does quite some work. Consider moving it into a separate method and add some comments describing the overall purpose and the non-obvious logic.

src/logging.cc
```diff
@@ -1148,13 +1221,50 @@ void LogFileObject::Write(bool force_flush,
...
+    CheckHistoryFileNum();
+    const char *filename = file_list_.back().name.c_str();
+    int fd = open(filename, O_WRONLY | O_CREAT | O_APPEND, FLAGS_logfile_mode);
+    if (fd != -1) {
+#ifdef HAVE_FCNTL
+      // Mark the file close-on-exec. We don't really care if this fails
+      fcntl(fd, F_SETFD, FD_CLOEXEC);
+#endif
+      file_ = fopen(filename, "a+"); // Read and append a FILE*.
+      if (file_ == NULL) {      // Man, we're screwed!, try to create new log file
```
sergiud, line 1260: When's/where's the new log file created as mentioned in the comment? This is not clear.

sergiud (changes requested):
Thanks for the PR. I have several remarks. Could you also please rebase?

src/logging.cc
yangzhg: done

src/logging.cc
yangzhg: done

src/logging.cc
yangzhg: done

src/logging.cc
yangzhg: done

src/logging.cc
yangzhg: Thanks for review, it is because of need to access the last element @ https://github.com/google/glog/blob/0f7ffec4ac55a5513ab10d85d2d51ac60e6bd4a9/src/logging.cc#L1252

src/logging.cc
yangzhg: done

src/logging.cc
yangzhg: done

yangzhg:
@sergiud I have edited according to your comments, please review again

src/logging.cc
sergiud: That's not a good reason for choosing `std::list`. Given you are continually sorting, `std::priority_queue` is probably a better choice.

sergiud:
Thanks! Unfortunately, Windows unit tests are failing now. Could you please take a look at these first?

yangzhg:
> Thanks! Unfortunately, Windows unit tests are failing now. Could you please take a look at these first?

Ok.  I will find a Windows device to test it in this weekend

src/logging.cc
yangzhg: Maybe `std::set` is a better choice? It is ordered and have access to first and last element.

sergiud: No, `std::set` is only useful if you lookup specific elements which you don't. The container also doesn't really give direct access to the top element. This is what `std::priority_queue` is for.

yangzhg:
> > Thanks! Unfortunately, Windows unit tests are failing now. Could you please take a look at these first?
> 
> Ok. I will find a Windows device to test it in this weekend

I have tested it on Windows vs 2022 and not failed now, would you please trigger the build again?
image

src/logging.cc
```diff
@@ -476,14 +497,24 @@ class LogFileObject : public base::Logger {
   uint32 bytes_since_flush_{0};
   uint32 dropped_mem_length_{0};
   uint32 file_length_{0};
-  unsigned int rollover_attempt_;
+  uint32 rollover_attempt_;
   int64 next_flush_time_{0};  // cycle count at which to flush log
   WallTime start_time_;
+  std::multiset<Filetime> file_list_;
```
sergiud, line 503: I'm not sure why a `std::multiset` is needed here. Can a file be associated with a multiple timestamps? I don't see how this is possible. Either way, a `multiset` would rather memory inefficient here, since the same file name would be potentially stored multiple times.

Please drop the `Filetime` structure and use an associative container. I guess a `std::map` is sufficient here.

sergiud (changes requested):
@yangzhg I've rebased your branch. Let's continue from here.

IceLemon99:
I just pull the branch code test roll file function, when restart test demo many times the file count will over  FLAGS_max_logfile_num. Is this a problem ?

sergiud:
@IceLemon99 I'll close the PR since it's heavily diverged from `master`. If you want continue working on this please rebase.

PR #829, opened 2022-06-06
simplify DLOG macro definition in NDEBUG mode

joliph:
Simplify the DLOG macro definition in NDEBUG mode with consistent logic


joliph:
retry

PR #848, opened 2022-08-12
cmake: disable install rules unless standalone build.

anpol:
If a library is embedded into an outer project, the library install rules should be disabled not to interfere with the outer project install rules.

A similar setting is used by other projects:

* https://github.com/abseil/abseil-cpp/blob/97ab3dcfd6490434202e4ab00b2eaba9449e42a1/CMakeLists.txt#L60-L64
* https://github.com/google/googletest/blob/7735334a46da480a749945c0f645155d90d73855/CMakeLists.txt#L32

For the same reason, we are unwelcome to modify a default value for a shared BUILD_SHARED_LIBS option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,15 @@ include (GenerateExportHeader)
 include (GetCacheVariables)
 include (GNUInstallDirs)
 
-option (BUILD_SHARED_LIBS "Build shared libraries" ON)
+# When glog is included as a subproject, do not apply standalone defaults.
+if (NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+  set(_glog_standalone OFF)
+else ()
+  set(_glog_standalone ON)
+  option (BUILD_SHARED_LIBS "Build shared libraries" ON)
+endif ()
+
+option(GLOG_ENABLE_INSTALL "glog: Enable install rules" ${_glog_standalone})
 option (PRINT_UNSYMBOLIZED_STACK_TRACES
   "Print file offsets in traces instead of symbolizing" OFF)
 option (WITH_CUSTOM_PREFIX "Enable support for user-generated message prefixes" ON)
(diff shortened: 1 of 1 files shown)


sergiud:
Generally, I'm not keen providing the extra logic just to support the niche case where glog is embedded per `add_subdirectory` into an unrelated project. In such case, there are additional gotchas that also must be handled to be fully robust. However, this introduces additional generally unnecessary and hard to maintain code.

My general advice is not to use `add_subdirectory` but to compartmentalize the unrelated CMake namespaces via `ExternalProject`. `FetchContent` may work as well.

anpol:
Thank you for your advice, I did't consider `ExternalProject` previously.

As far as I understand, CMake's `ExternalProject` is a tool that isolates a build of a library from a build of an outer project. So all configuration, build, install, test steps are totally isolated, they could even be performed by a completely different toolchain. That could lead to various kinds of inconsistency. And I suspect won't get the sources of an external project visible in IDEs like Visual Studio or Xcode, right? Looks like ExternalProject is a tool designed to solve a different problem under stronger constraints.

If I were using `ExternalProject_Add`, I would rather build glog with Bazel, ignoring glog's CMakeLists.txt file.

For now, I would still prefer using `add_subdirectory` [if it works](https://cmake.org/pipermail/cmake/2013-February/053429.html), as it allows to glue CMake projects seamlessly.

I will understand if you keep glog's CMakeLists.txt for standalone build only, just to ease maintenance burden. CMake scripts are hard to maintain, so it's better to keep them simple.  That is exactly what I'm trying to achieve, but from a user's perspective.

sergiud:
> As far as I understand, CMake's ExternalProject is a tool that isolates a build of a library from a build of an outer project. 

Correct, and this is what you are trying to achieve with this PR (albeit only partially): to control certain options independently from the outer project.

> So all configuration, build, install, test steps are totally isolated, they could even be performed by a completely different toolchain. That could lead to various kinds of inconsistency. 

The resulting flexibility theoretically allows the use of a different toolchain, yes. However, this is an atypical use case and not the way the module is usually employed.

I would not agree that inconsistency between unrelated projects is likely to cause actual problems unless you have specific problems in mind. I'd rather argue that disassociating options from their intended use generates an inconsistency that is to be avoided.

> For now, I would still prefer using add_subdirectory [if it works](https://cmake.org/pipermail/cmake/2013-February/053429.html), as it allows to glue CMake projects seamlessly.

You can include glog via `add_subdirectly` already today. Evidently, specific requirements such as yours may not be well supported. However, given CMake already provides specific functionality that allows to overcome these limitations it would be unfair to shift the maintenance burden on glog.

PR #849, opened 2022-08-12
windows: call to FormatMessage() should _IGNORE_INSERTS

anpol:
Calling FormatMessage without a FORMAT_MESSAGE_IGNORE_INSERTS flag could cause trouble, as explained in an article by Raymond Chen:

* The importance of the FORMAT_MESSAGE_IGNORE_INSERTS flag
  https://devblogs.microsoft.com/oldnewthing/20071128-00/?p=24353


sergiud:
Sounds reasonable. Thanks!

PR #850, opened 2022-08-12
log_file_header: add option to disable log file headers.

anpol:
Log lines can be customized for parsing by an external tool.

To simplify such customization and parsing, there should be an option to customize the file header, or at least to disable adding it.


sergiud:
Thank you!

PR #851, opened 2022-08-12
Fix broken stacktrace

marekcirkos:
Since #846, `HAVE_UNWIND_H`  is not really in use. Instead we should use `HAVE__UNWIND_BACKTRACE` and `HAVE__UNWIND_GETIP` (added in #846).
To prevent that from happening again, also added Bazel tests that confirm stacktrace are still working.


marekcirkos:
@drigz are you windows guy by any chance?

test/stacktrace/test.cpp
```diff
@@ -0,0 +1,62 @@
+#include <algorithm>
```
sergiud, line 1: Are you sure this is necessary? There's a `stacktrace_unittest.cc` already. Either way, the naming scheme does not follow glog conventions.

marekcirkos:
Thanks @sergiud I missed existing tests. I added Bazel rules for them, skipped few however, as I didn't have time to investigate why they are broken.

There is some caveat, though. Current `symbolize_test` and `stacktrace_test` will miss this issue anyway. Mainly because they will just "skip" due to missing `HAVE_SYMBOLIZE` or/and `HAVE_STACKTRACE` definitions eg. https://github.com/google/glog/blob/master/src/stacktrace_unittest.cc#L236.

From what I can see they are defined in here https://github.com/google/glog/blob/master/src/utilities.h#L104.

We could define those flags in Bazel, but that might fail tests in legit platforms that do not support them.

drigz:
> @drigz are you windows guy by any chance?

I'm afraid not, almost all my Windows development experience comes from trying to keep this build working on Windows... Why do you ask?

If it's about the current error:

```
c:\b\k4x6jequ\execroot\__main__\src\utilities.h(108): error C2220: warning treated as error - no 'object' file generated
--
  | c:\b\k4x6jequ\execroot\__main__\src\utilities.h(108): warning C4005: 'HAVE_STACKTRACE': macro redefinition
  | src/stacktrace_unittest.cc: note: see previous definition of 'HAVE_STACKTRACE'
```

I hope that `/IGNORE:C4005` could help ([IGNORE](https://docs.microsoft.com/en-us/cpp/build/reference/ignore-ignore-specific-warnings?view=msvc-170), [C4005](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4005?view=msvc-170)).

However, if this is too fiddly, we could also consider a different approach: a bazel-only test that runs `EXPECT_TRUE(HAVE_STACKTRACE)`. I guess that the bazel tests only run on CI, and we do expect stacktrace support on Linux/macOS/Windows - so hopefully we don't need to worry about platforms that truly don't support stacktraces.

That said, I'd accept this PR even without tests, as adding a new test just to check we've kept `copts` up to date feels like scope creep.

Thanks again for taking the time to get to the bottom of this!

marekcirkos:
@drigz adjusted test per your suggestion :)

drigz:
Bazel changes LGTM, @sergiud if you're happy with the change to stacktrace_unittest.cc please submit.

sergiud:
Thanks!

PR #853, opened 2022-08-13
SetLogger should delete previously set custom logger.

anpol:
As specified in the doc comment for `SetLogger`, _"the logger becomes the property of the logging module and should not be deleted by the caller"_.

Not only should the `LogDestination` delete a custom logger in its destructor, but it should also delete a previous logger when another logger is passed to `SetLogger()`.

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -649,7 +649,10 @@ LogDestination::~LogDestination() {
 }
 
 void LogDestination::SetLoggerImpl(base::Logger* logger) {
-  if (logger_ == logger) return;
+  if (logger_ == logger) {
+    // Prevent releasing currently held sink on reset
+    return;
+  }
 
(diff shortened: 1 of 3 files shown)


src/logging.cc
```diff
@@ -641,10 +645,17 @@ LogDestination::LogDestination(LogSeverity severity,
 }
 
 LogDestination::~LogDestination() {
+  ResetLoggerImpl();
+}
+
+void LogDestination::SetLoggerImpl(base::Logger* logger) {
+  if (logger_ == logger) return;
```
sergiud, line 652:
```suggestion
  if (logger_ == logger) {
    // Prevent releasing currently held sink on reset
    return;
  }
```

sergiud (changes requested):
Thank you! I have only a minor remark.

sergiud (approved):

Issue #859, opened 2022-09-08
symbolize_test fails on Bazel CI (Windows only)


Wyverald:
Example breakage: https://buildkite.com/bazel/google-logging/builds/2911#01831a72-99a7-4d0a-a1a6-4761066ca927

This has been consistently failing, with both HEAD Bazel and latest release Bazel (5.3.0).

drigz:
Thanks for the report, and let me know if there's something I should do to opt-in to automated notifications of breakage on master. I don't send PRs regularly enough to see the breakage there.

- last green: [build 2894](https://buildkite.com/bazel/google-logging/builds/2894#0182c2de-b916-4e0c-83a5-d02f2ab9ba84) 22nd August with bazel 5.2.0 at [b70ea80](https://github.com/google/glog/commit/b70ea80433c2a8f20b832be97b90f1f82b0d29e9)
- first red: [build 2895](https://buildkite.com/bazel/google-logging/builds/2895#0182c819-b174-4f07-ad1b-1c899e8c75db) 23rd August with bazel 5.2.0 at [b70ea80](https://github.com/google/glog/commit/b70ea80433c2a8f20b832be97b90f1f82b0d29e9)

Hard to guess what might have changed to cause this. Test artifacts contain:

```
Executing tests from //:symbolize_test
-----------------------------------------------------------------------------
F20220908 00:14:51.232890  3688 symbolize_unittest.cc:431] Check failed: symbol != NULL 
*** Check failure stack trace: ***
    @   00007FF7A7B1B9BA  ?Fail@LogMessage@google@@SAXXZ
    @   00007FF7A7B1B8D0  ?SendToLog@LogMessage@google@@QEAAXXZ
    @   00007FF7A7B1B0B0  ?Flush@LogMessage@google@@QEAAXXZ
    @   00007FF7A7B1CC6C  ??1LogMessageFatal@google@@QEAA@XZ
    @   00007FF7A7B14074  (unknown)
    @   00007FF7A7B1417F  (unknown)
    @   00007FF7A7B4CFF0  ?InstallFailureWriter@google@@YAXP6AXPEBD_K@Z@Z
    @   00007FFE37EB7974  BaseThreadInitThunk
    @   00007FFE3A56A2F1  RtlUserThreadStart
```

[...]

PR #860, opened 2022-09-13
Remove symbolize Bazel test

drigz:
It is broken on Windows CI (#859) in a way that I don't understand.


meteorcloudy:
Can we get this merged so that we can get a green CI 😃

PR #871, opened 2022-11-07
Provide API for logging with per-thread counters

boscosiu:
This is an RFC for a feature to allow counting loggers to use thread-local variables.

I'm writing task-based routines that spawn threads to repeatedly run a function or lambda.  Currently only the thread that "wins" the shared count to `N` gets to log which is suboptimal for this use case.

It would be nice to be able to periodically report progress from each thread with independent counters for each task's loop.

If this is something we'd want to add, I'll follow up with docs and tests.  Thanks for looking!


sergiud:
Thanks for the PR. Given there are is no feedback and the PR heavily diverged from `master` we can abandon the changes. In case you want continue working on this, you can rebase anytime.

PR #877, opened 2022-11-29
Fix Windows clang CI for CC toolchain resolution

### LINKED ISSUE #876: Flag --incompatible_enable_cc_toolchain_resolution will break Google Logging in Bazel 7.0
Incompatible flag `--incompatible_enable_cc_toolchain_resolution` will be enabled by default in the next major release (Bazel 7.0), thus breaking Google Logging. Please migrate to fix this and unblock the flip of this flag.

The flag is documented here: https://github.com/bazelbuild/bazel/issues/7260

Please check the following CI builds for build and test results:

- [Windows, Clang, OpenJDK 11](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1339#018497b7-aa0d-448d-8747-216101a78eb8)

Never heard of incompatible flags before? We have [documentation](https://docs.bazel.build/versions/master/backward-compatibility.html) that explains everything.
If you have any questions, please file an issue in https://github.com/bazelbuild/continuous-integration.

jsharpe:
Select the correct toolchain and platform for
building with clang on windows with CC toolchain
resolution active.

Attempting to fix #876

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -47,13 +47,15 @@ tasks:
     environment:
       BAZEL_VC: "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\VC"
     build_flags:
-    - "--noincompatible_enable_cc_toolchain_resolution"
+    - "--extra_toolchains=@local_config_cc//:cc-toolchain-x64_windows-clang-cl"
+    - "--extra_execution_platforms=//:x64_windows-clang-cl"
     - "--compiler=clang-cl"
     - "--features=layering_check"
     build_targets:
     - "//..."
     test_flags:
-    - "--noincompatible_enable_cc_toolchain_resolution"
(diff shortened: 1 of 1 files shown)


jsharpe:
@comius @meteorcloudy I think this should fix the incompatible_enable_cc_toolchain_resolution build for this package.

drigz:
Thanks, buildkite logs look good! Do you know if `--incompatible_enable_cc_toolchain_resolution` is already the default at head, or should we run a build explicitly setting `--incompatible_enable_cc_toolchain_resolution` to check this?

comius:
> Do you know if --incompatible_enable_cc_toolchain_resolution is already the default at head, or should we run a build explicitly setting --incompatible_enable_cc_toolchain_resolution to check this?

It's false at head, you need to set it manually.

drigz:
Many thanks, Ivo. I explicitly enabled it for a test build and it seems happy: https://buildkite.com/bazel/google-logging/builds/3010#0184c7ce-6aa4-4803-8e30-50ab4834352f

PR #878, opened 2022-12-08
Adds fuzz testing for demangle

catenacyber:
@sergiud here is a PR to get fuzzing built with cmake
cf https://github.com/google/oss-fuzz/pull/9160#issuecomment-1341358902

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,9 @@ option (WITH_UNWIND "Enable libunwind support" ON)
 
 cmake_dependent_option (WITH_GMOCK "Use Google Mock" ON WITH_GTEST OFF)
 
+set (WITH_FUZZING none CACHE STRING "Fuzzing engine")
+set_property (CACHE WITH_FUZZING PROPERTY STRINGS none libfuzzer ossfuzz)
+
 if (NOT WITH_UNWIND)
   set (CMAKE_DISABLE_FIND_PACKAGE_Unwind ON)
 endif (NOT WITH_UNWIND)
@@ -747,6 +750,22 @@ if (WITH_PKGCONFIG)
 endif (WITH_PKGCONFIG)
 
 # Unit testing
+
+if (NOT WITH_FUZZING STREQUAL "none")
+  add_executable (fuzz_demangle
+    src/fuzz_demangle.cc
+  )
+
+  if (WITH_FUZZING STREQUAL "ossfuzz")
+    set (LIB_FUZZING_ENGINE $ENV{LIB_FUZZING_ENGINE})
+    target_link_libraries (fuzz_demangle PRIVATE glog ${LIB_FUZZING_ENGINE})
+  elseif (WITH_FUZZING STREQUAL "libfuzzer")
+    target_compile_options (fuzz_demangle PRIVATE -fsanitize=fuzzer)
+    target_link_libraries (fuzz_demangle PRIVATE glog)
+  else (WITH_FUZZING STREQUAL "libfuzzer")
+    message (FATAL_ERROR "Unsupported fuzzing engine ${WITH_FUZZING}")
+  endif (WITH_FUZZING STREQUAL "ossfuzz")
+endif (NOT WITH_FUZZING STREQUAL "none")
 
 if (BUILD_TESTING)
   add_library (glogtest STATIC
--- a/src/fuzz_demangle.cc
+++ b/src/fuzz_demangle.cc
@@ -1,0 +1,32 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////////
+
+#include <cstring>
+
+#include "demangle.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const unsigned char *Data,
+                                      unsigned Size) {
+  if (Size >= 4095) {
+    return 0;
+  }
+  char Buffer[Size + 1];
+  std::memcpy(Buffer, Data, Size);
+  Buffer[Size] = 0;
+  char demangled[4096];
+  google::Demangle(Buffer, demangled, Size);
+  return 0;
+}


CMakeLists.txt
```diff
@@ -743,6 +743,13 @@ endif (WITH_PKGCONFIG)
 
 # Unit testing
 
+if (BUILD_FUZZING)
+  add_executable (fuzz_demangle
+    src/fuzz_demangle.cc
+  )
+  target_link_libraries(fuzz_demangle glog $ENV{LIB_FUZZING_ENGINE})
+endif (BUILD_FUZZING)
```
sergiud, line 751: I would use a CMake variable here. The caller can still use an environment variable if he chooses to do so.
```suggestion
if (BUILD_FUZZING AND WITH_FUZZING_ENGINE)
  add_executable (fuzz_demangle
    src/fuzz_demangle.cc
  )
  target_link_libraries (fuzz_demangle PRIVATE glog ${WITH_FUZZING_ENGINE})
endif (BUILD_FUZZING AND WITH_FUZZING_ENGINE)
```
```suggestion
  target_link_libraries (fuzz_demangle PRIVATE glog ${WITH_FUZZING_ENGINE})
```
`WITH_FUZZING_ENGINE` should be populated with sensible choices including one that disables fuzzing altogether (e.g., `NO` could work) which should also probably be the default.

src/fuzz_demangle.cc
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////////
+
+#include "demangle.h"
+
+using namespace GOOGLE_NAMESPACE;
+
+extern "C" int LLVMFuzzerTestOneInput(const unsigned char *Data, unsigned Size) {
```
sergiud, line 21: Why `LLVM` here? Is the naming scheme prescribed by the fuzzing engine? If so, can we use different engines at all as the CMake setup suggests?

src/fuzz_demangle.cc
```diff
@@ -0,0 +1,31 @@
...
+  if (Size >= 4095) {
+    return 0;
+  }
+  char Buffer[Size + 1];
+  memcpy(Buffer, Data, Size);
```
sergiud, line 26:
```suggestion
  std::memcpy(Buffer, Data, Size);
```
IWYU: missing `#include <cstring>` at the top.

src/fuzz_demangle.cc
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////////
+
+#include "demangle.h"
+
+using namespace GOOGLE_NAMESPACE;
```
sergiud, line 19:
```suggestion
```
Not necessary, as `Demangle` is explicitly referenced below.

sergiud (changes requested):
Sorry for the late review. I have a couple of remarks that need to be addressed.

src/fuzz_demangle.cc
catenacyber: ok

src/fuzz_demangle.cc
catenacyber: ok

src/fuzz_demangle.cc
catenacyber: > Why `LLVM` here? Is the naming scheme prescribed by the fuzzing engine?

The name is indeed prescribed cf https://llvm.org/docs/LibFuzzer.html#id23

>  If so, can we use different engines at all as the CMake setup suggests?

oss-fuzz uses different fuzzing engines : libfuzzer, afl, honggfuzz... It manages to do so with every fuzz target having `LLVMFuzzerTestOneInput` as its entry point. If I am correct, to use other engines, there are simply wrappers that call `LLVMFuzzerTestOneInput`...

CMakeLists.txt
catenacyber: Good idea for the cmake variable `WITH_FUZZING_ENGINE`.

I am not sure I understand what you expect by `should be populated with sensible choices` (maybe I do not know cmake enough)

catenacyber:
friendly ping @sergiud ;-)

sergiud:
Thanks for the ping @catenacyber. I'm very sorry for the delay. I will look into PR during the coming week. Please stay tuned.

sergiud:
Ideally, we would run fuzzing as part of our Github workflow. What do we need for that?

sergiud (changes requested):

CMakeLists.txt
sergiud: See my additional comments.

src/fuzz_demangle.cc
```diff
@@ -0,0 +1,30 @@
+// Copyright 2021 Google LLC
```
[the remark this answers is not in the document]
catenacyber: fixing

CMakeLists.txt
```diff
@@ -743,6 +743,13 @@ endif (WITH_PKGCONFIG)
 
 # Unit testing
 
+if (BUILD_FUZZING AND WITH_FUZZING_ENGINE)
+  add_executable (fuzz_demangle
+    src/fuzz_demangle.cc
+  )
+  target_link_libraries(fuzz_demangle glog ${WITH_FUZZING_ENGINE})
```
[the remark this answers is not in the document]
catenacyber: fixing

CMakeLists.txt
```diff
@@ -743,6 +743,13 @@ endif (WITH_PKGCONFIG)
 
 # Unit testing
 
+if (BUILD_FUZZING AND WITH_FUZZING_ENGINE)
```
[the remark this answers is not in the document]
catenacyber: Trying something with only option `WITH_FUZZING`...

What do you expect for a `none` value ? Do not build fuzzers ? Build a static library of the fuzz target ? Have a simple driver like https://github.com/OISF/suricata/blob/master/src/tests/fuzz/onefile.c ?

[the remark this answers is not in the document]
sergiud: `none` simply disables fuzzing altogether.

CMakeLists.txt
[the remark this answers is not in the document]
catenacyber: The INTERFACE, PUBLIC or PRIVATE option must appear as the second argument, just after the target name.

CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ option (WITH_TLS "Enable Thread Local Storage (TLS) support" ON)
 option (WITH_UNWIND "Enable libunwind support" ON)
 
 cmake_dependent_option (WITH_GMOCK "Use Google Mock" ON WITH_GTEST OFF)
+set (WITH_FUZZING -fsanitize=fuzzer CACHE STRING "Fuzzing engine")
```
sergiud, line 50: The option is not meant to hold compiler flags or the like. It should be a symbolic name (e.g., fuzzer). Depending on the choice the corresponding flags or link libraries should be added. Something like

```cmake
if (WITH_FUZZING STREQUAL fuzzer)
  target_compile_options (fuzz_demangle PRIVATE -fsanitize=fuzzer)
endifif (WITH_FUZZING STREQUAL fuzzer)
```

CMakeLists.txt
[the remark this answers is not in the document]
sergiud: Yes, correct.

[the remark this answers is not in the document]
catenacyber: fixed

CMakeLists.txt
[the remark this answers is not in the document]
catenacyber: `none` works as expected now

CMakeLists.txt
catenacyber: Ok @sergiud 

Should be as you expect now ;-)

CMakeLists.txt
```diff
@@ -745,10 +746,16 @@ endif (WITH_PKGCONFIG)
...
-  add_executable (fuzz_demangle
+  add_executable (fuzz_detangle
     src/fuzz_demangle.cc
   )
-  target_link_libraries (fuzz_demangle PRIVATE glog ${WITH_FUZZING})
+  if (WITH_FUZZING STREQUAL "ossfuzz")
+    target_link_libraries (fuzz_demangle PRIVATE glog $ENV{LIB_FUZZING_ENGINE})
+  elseif (WITH_FUZZING STREQUAL "libfuzzer")
+    target_link_libraries (fuzz_demangle PRIVATE glog -fsanitize=fuzzer)
+  else
```
sergiud: Incorrect syntax:
```suggestion
  else (WITH_FUZZING STREQUAL "libfuzzer")
```

CMakeLists.txt
```diff
@@ -745,10 +746,16 @@ endif (WITH_PKGCONFIG)
...
 if (WITH_FUZZING AND NOT WITH_FUZZING STREQUAL "none")
-  add_executable (fuzz_demangle
+  add_executable (fuzz_detangle
     src/fuzz_demangle.cc
   )
-  target_link_libraries (fuzz_demangle PRIVATE glog ${WITH_FUZZING})
+  if (WITH_FUZZING STREQUAL "ossfuzz")
+    target_link_libraries (fuzz_demangle PRIVATE glog $ENV{LIB_FUZZING_ENGINE})
+  elseif (WITH_FUZZING STREQUAL "libfuzzer")
+    target_link_libraries (fuzz_demangle PRIVATE glog -fsanitize=fuzzer)
```
sergiud, line 755:
```suggestion
    target_compile_options (fuzz_demangle PRIVATE -fsanitize=fuzzer)
    target_link_libraries (fuzz_demangle PRIVATE glog)
```

CMakeLists.txt
```diff
@@ -745,10 +746,16 @@ endif (WITH_PKGCONFIG)
 # Unit testing
 
 if (WITH_FUZZING AND NOT WITH_FUZZING STREQUAL "none")
-  add_executable (fuzz_demangle
+  add_executable (fuzz_detangle
     src/fuzz_demangle.cc
   )
-  target_link_libraries (fuzz_demangle PRIVATE glog ${WITH_FUZZING})
+  if (WITH_FUZZING STREQUAL "ossfuzz")
+    target_link_libraries (fuzz_demangle PRIVATE glog $ENV{LIB_FUZZING_ENGINE})
```
sergiud, line 753: Why do we need `LIB_FUZZING_ENGINE` here? Isn't the `WITH_FUZZING` value sufficient? Either way, I would avoid using an environment variable here and instead use a CMake variable directly (drop `$ENV`). One can always initialize a CMake variable using an environment variable which is the usual pattern.

CMakeLists.txt
catenacyber: ok

CMakeLists.txt
catenacyber: ok

CMakeLists.txt
catenacyber: oss-fuzz uses this environment variable to use whatever engine such as libfuzzer, afl++ or honggfuzz or other ones...

As you said 

> The option is not meant to hold compiler flags or the like. It should be a symbolic name (e.g., fuzzer)

So, I chose ossfuzz for this setup.
What do you think about it ?

So, I divide in 2 lines : setting the cmake variable from the environment and then using it...

sergiud: Shouldn't we split `ossfuzz` into libfuzzer, afl++ or honggfuzz then? This would provide a much better overview of the choices we have.

catenacyber: I do not think so because oss-fuzz does not use regular afl++

And oss-fuzz is adding more engines, you just keep the env variable `LIB_FUZZING_ENGINE`, and get these for free

catenacyber:
@sergiud I hope I got cmake right ;-)

CMakeLists.txt
```diff
@@ -743,6 +745,21 @@ endif (WITH_PKGCONFIG)
...
+    src/fuzz_demangle.cc
+  )
+  if (WITH_FUZZING STREQUAL "ossfuzz")
+    set (LIB_FUZZING_ENGINE $ENV{LIB_FUZZING_ENGINE})
+    target_link_libraries (fuzz_demangle PRIVATE glog ${LIB_FUZZING_ENGINE})
+  elseif (WITH_FUZZING STREQUAL "libfuzzer")
+    target_compile_options (fuzz_demangle PRIVATE -fsanitize=fuzzer)
+    target_link_libraries (fuzz_demangle PRIVATE glog)
+  else(WITH_FUZZING STREQUAL "libfuzzer")
+    message( FATAL_ERROR "Unsupported fuzzing engine." )
```
sergiud, line 759:
```suggestion
    message (FATAL_ERROR "Unsupported fuzzing engine ${WITH_FUZZING}")
```

sergiud:
@catenacyber Almost there.

Another question: glog will be moving to C++14 shortly. Anything that we need to consider here?

CMakeLists.txt
sergiud: I see. Let's then keep it this way.

catenacyber:
I do not see why c++14 would be a problem.

sergiud:
OK, great!

One more thing: once you are done with the last change, please also rebase onto current master and squash if possible.

catenacyber:
> OK, great!
> 
> One more thing: once you are done with the last change, please also rebase onto current master and squash if possible.

Done ;-)

sergiud (approved):

sergiud:
Thanks!

Issue #837, opened 2022-07-06
bazel build does not mark includes as system headers


joshuarubin:
By not using [`includes`](https://docs.bazel.build/versions/main/be/c-cpp.html#cc_library.includes) in the `cc_library` for the bazel build, includes are considered local. This poses a problem for tools like clang-tidy that can suppress issues coming from system headers. As a result, our lint output is littered with glog errors.

drigz:
https://github.com/google/glog/commit/e6e2e1372aeb3f6ac6ca5f4a3d72bad8dce3c724 introduced strip_include_prefix. It's a "grab bag" commit, but I believe the relevant explanation is "No longer expose internal headers." Feel free to send a pull request to switch back, but please ensure that downstream targets that do eg `#include <stacktrace.h>` do not pick up the glog-internal header, since I think this is why the switch was made.

sergiud:
Is this still an issue?

drigz:
I guess so, I'm not aware of a way to avoid the "exposing the private config.h header" without having the effect that the reporter described, although presumably it could be done by rearranging/renaming the source/include files in this repo and adjusting the CMake build accordingly.

PR #880, opened 2022-12-08
Set includes for glog headers in Bazel library

sfc-gh-ebrossard:
Resolves https://github.com/google/glog/issues/837

This adds the `includes` property to the `cc_library` for `glog` so that includes are treated as system headers, suppressing issues related to lint, clang-tidy, etc. that may differ from the local project and `glog`.

### DIFF
--- a/bazel/glog.bzl
+++ b/bazel/glog.bzl
@@ -190,6 +190,7 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
             ":stl_logging_h",
             ":vlog_is_on_h",
         ],
+        includes = ["."],
         strip_include_prefix = "src",
         defines = final_lib_defines,
         copts = final_lib_copts,


sfc-gh-ebrossard:
I'm not sure it's feasible for this change to satisfy https://github.com/google/glog/issues/837#issuecomment-1178653000 or the failing tests, actually, given that `includes` has to specify a list of directories, and some of the headers specified in `hdrs` such as `logging.h` are in the "top-level" `glog` directory. The error is:
```
in includes attribute of cc_library rule //:glog: '.' resolves to the workspace root, which would allow this rule and all of its transitive dependents to include any file in your workspace. Please include only what you need.
```

drigz:
I'm afraid I've never fully understood the subtleties and the right way to hold `includes` vs `strip_include_prefix` but I feel this is not the right way to do it. Especially since exposing the internal files can break downstream projects if they include the wrong file by accident (this happened with `config.h`). My guess is that to fix #837 we would need to rearrange the header sources (eg add a top-level `include/`) and adapt the CMake build correspondingly, but I haven't looked into the details. Sorry :(

sfc-gh-ebrossard:
Yes, I think you're right. It would end up being much more invasive than simply setting `includes`, and with greater risk for breaking downstream projects.

It's probably more straightforward just to resolve the compiler warnings/clang-tidy issues/etc. that necessitate setting `includes` for some projects. There's also https://github.com/bazelbuild/rules_foreign_cc/issues/418, though, which is another can of worms...

I'll go ahead and close this PR, given that it doesn't seem like the right approach in any case.

PR #890, opened 2023-02-06
Add tsan annotation to static variable when checking if vlog is on.

pateldeev:
The `vlocal__` static variable triggers tsan when multiple threads try and see if vlog is on.


Miniiii1989 (approved):

sergiud:
Thanks!

PR #894, opened 2023-02-18
Change SleepForMilliseconds parameter from unsigned int -> long

kraj:
Fixes following error on 32bit platforms

src/googletest.h:631:35: error: non-constant-expression cannot be narrowed from type 'unsigned int' to 'long' in initializer list [-Wc++11-narrowing]
  const struct timespec req = {0, t * 1000 * 1000};
                                  ^~~~~~~~~~~~~~~

### DIFF
--- a/src/googletest.h
+++ b/src/googletest.h
@@ -625,7 +625,7 @@ class Thread {
 #endif
 };
 
-static inline void SleepForMilliseconds(unsigned t) {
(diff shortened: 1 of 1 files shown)


sergiud:
Thanks. However, your change breaks multiple builds.

sergiud:
Let us know when you are ready to continue working on the PR.

PR #898, opened 2023-02-24
drop custom prefix option

sergiud:
We already enable custom prefix by default (#788). Promote custom prefix to first class citizen in order to eliminate unit test code duplication.

cc @santigl


sergiud:
@santigl Good point, thanks. Updated.

PR #903, opened 2023-02-27
ci: includ CIFuzz GitHub workflow

catenacyber:
@sergiud 

Looks like oss-fuzz is working well and generating reports, following https://github.com/google/glog/pull/878
This is the PR fo get fuzzing run for every next PR

But I advise to fix the existing reports first


sergiud:
Thank you!

Issue #904, opened 2023-02-28
Stack trace is empty when use bazel on aarch64


noaxp:
Platform: jetson-orin linux aarch64
Compiler: bazel
```
*** Check failure stack trace: ***
*** Aborted at 1677576331 (unix time) try "date -d @1677576331" if you are using GNU date ***
PC: @                0x0 (unknown)

```

The root cause is here https://github.com/google/glog/blob/master/src/utilities.h#L103
1. Jetson orin's arch is aarch64, so we expect to use `stacktrace_generic-inl.h`
2. But we didn't define `HAVE_EXECINFO_BACKTRACE` in `glog.bazel`

I try to add `-DHAVE_EXECINFO_BACKTRACE` to `linux_or_darwin_copts`
```
--- a/bazel/glog.bzl
+++ b/bazel/glog.bzl
@@ -81,6 +81,7 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
         "-DHAVE_SYS_SYSCALL_H",
         # For src/logging.cc to create symlinks.
         "-DHAVE_UNISTD_H",
+        "-DHAVE_EXECINFO_H",
     ]
```
Then the issue is solved.

So I think we should add `linux_only_copts = ["-DHAVE_EXECINFO_H"]` to `glog.bzl`.
I will be happy to fix it if you confirm this solution.

drigz:
Please send a PR, it seems reasonable to expect execinfo.h on Linux.

PR #905, opened 2023-02-28
Fix empty stack trace when use bazel on aarch64

noaxp:
Fix issue https://github.com/google/glog/issues/904: Stack trace is empty when use bazel on aarch64

Add `-DHAVE_EXECINFO_H` to linux copts.


drigz:
I guess we can ignore the failed checks, they seem like flakes.

drigz:
Thank you!

PR #916, opened 2023-04-24
Fix targets for layering_check

mai93:
Including headers that are not public in the dependencies broke the build with Bazel@Head because it violates layering_check [error](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2975#0187b16f-1eaf-4da2-994c-480706d4848e). This PR moves headers required by the test targets to `hdrs` attribute of their direct dependencies.


mai93:
Thanks @drigz! That makes sense. I think we can move the shared headers to a filegroup and have the `glog` and test targets use it.

PR #925, opened 2023-05-25
logging: Prevent LogStream constructor from being discarded

### LINKED ISSUE #922: Linker error 'construction vtable ... LogMessage::LogStream'
Hello, I am encountering a very weird issue when linking a clang-built executable to a GCC-built glog static library on Ubuntu 22.04.

I've reduced the issue to a simple example

```cpp
#define GOOGLE_STRIP_LOG 1

#include <glog/logging.h>

int main(int argc, char* argv[]) {
    // Initialize Google’s logging library.

    VLOG(2) << "stuffs";
}
```

I get the following error:

```
[1/2] /usr/bin/clang++ -DGFLAGS_IS_A_DLL=0 -isystem /workspace/build/glog-gcc-prefix/include -O3 -DNDEBUG -MD -MT CMakeFiles/issue.dir/main.cpp.o -MF CMakeFiles/issue.dir/main.cpp.o.d -o CMakeFiles/issue.dir/main.cpp.o -c /workspace/repro/main.cpp
[2/2] : && /usr/bin/clang++ -O3 -DNDEBUG  CMakeFiles/issue.dir/main.cpp.o -o issue  /workspace/build/glog-gcc-prefix/lib/libglog.a  /usr/lib/x86_64-linux-gnu/libgflags.so.2.2.2  -lpthread && :
FAILED: issue
: && /usr/bin/clang++ -O3 -DNDEBUG  CMakeFiles/issue.dir/main.cpp.o -o issue  /workspace/build/glog-gcc-prefix/lib/libglog.a  /usr/lib/x86_64-linux-gnu/libgflags.so.2.2.2  -lpthread && :
/usr/bin/ld: /workspace/build/glog-gcc-prefix/lib/libglog.a(logging.cc.o): relocation R_X86_64_PC32 against undefined hidden symbol `_ZTCN6google10LogMessage9LogStreamE0_So' can not be used when making a PIE object
/usr/bin/ld: final link failed: bad value
clang: error: linker command failed with exit code 1 (use -v to see invocation
```

Using `c++filt` to demangle:

```
$ c++filt _ZTCN6google10LogMessage9LogStreamE0_So
construction vtable for std::basic_ostream<char, std::char_traits<char> >-in-google::LogMessage::LogStream
```

If I remove `#define GOOGLE_STRIP_LOG 1` it will compile, but I don't think this is a good solution, unless there's something I should be more aware about when using `GOOGLE_STRIP_LOG`.

Moreover, after some digging, I found that the file will compile with _no optimizations_ and fail with optimizations enabled (`-O1`, `-O2`, `-O3`), meaning that a CMake `Debug` build will succeed but a `Release` build will fail.

I have a repo to help reproduce the issue here https://github.com/ekilmer/glog-linker-repro-issue and it also has more details and instructions.

Happy to provide more details, if needed.

ekilmer:
Fixes linker error reported in https://github.com/google/glog/issues/922

closes #922

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -29,6 +29,7 @@ Andy Ying <andy@trailofbits.com>
 Bret McKee <bretmckee@google.com>
 Brian Silverman <bsilver16384@gmail.com>
 Dmitriy Arbitman <d.arbitman@gmail.com>
+Eric Kilmer <eric.d.kilmer@gmail.com>
 Fumitoshi Ukai <ukai@google.com>
 Guillaume Dumont <dumont.guillaume@gmail.com>
 Håkan L. S. Younes <hyounes@google.com>
--- a/src/glog/logging.h.in
+++ b/src/glog/logging.h.in
@@ -1344,6 +1344,15 @@ GLOG_MSVC_PUSH_DISABLE_WARNING(4275)
   class GLOG_EXPORT LogStream : public std::ostream {
 GLOG_MSVC_POP_WARNING()
   public:
+#if defined __has_attribute
+#  if __has_attribute (used)
+    // In some cases, like when compiling glog as a static library with GCC and
+    // linking against a Clang-built executable, this constructor will be
+    // removed by the linker. We use this attribute to prevent the linker from
+    // discarding it.
+    __attribute__ ((used))
+#  endif
+#endif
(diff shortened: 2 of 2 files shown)


sergiud:
Thanks!

PR #926, opened 2023-05-28
Switch strip_include_prefix -> includes

cpsauer:
Hey Googlers!

I was helping a Googler use a bazel tool I'd written over in https://github.com/hedronvision/bazel-compile-commands-extractor/issues/121 and noticed that the BUILD file used a `strip_include_prefix` where an `includes` will do. Swiching to `includes` would save some symlinking (see https://github.com/bazelbuild/bazel/issues/17591) and probably the windows pain you guys were experiencing there. 

So I figured toss up a quick PR to change it over, fixing his issue and trying to leave things better than I found them.  Second commit coming in a sec to propose unwinding the windows hack, now that `strip_include_prefix` is no longer in play.

Cheers,
Chris
(ex-Googler and author of some Bazel tools)

### DIFF
--- a/bazel/glog.bzl
+++ b/bazel/glog.bzl
@@ -204,10 +204,7 @@ def glog_library(namespace = "google", with_gflags = 1, **kwargs):
         includes = ["src"],
         defines = final_lib_defines,
         copts = final_lib_copts,
-        deps = gflags_deps + select({
-            "@bazel_tools//src/conditions:windows": [":strip_include_prefix_hack"],
-            "//conditions:default": [],
-        }),
+        deps = gflags_deps,
         **kwargs
     )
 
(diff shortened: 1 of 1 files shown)


drigz:
Hey Chris, thanks for sending this. We should add a comment but strip_include_prefix is there for a reason: https://github.com/google/glog/issues/837#issuecomment-1178653000 When setting up these build rules we found a bunch of cases where Bazel exposed "private" headers to downstream targets and I'm reluctant to change anything here as my understand has only gotten rustier since then.

As far as I'm aware, this change will mean that downstream targets that use `#include <stacktrace.h>` might pick up glog's internal header rather than their own headers as they expect. Unfortunately, I'm not aware of a good way to test for this (until thinking about it just now, I'd thought we'd need an external repository, although maybe we could have a test within this repo that does `#ifdef #error` for a header defined in stacktrace.h...). Can you check whether this PR will change that behavior?

cpsauer:
Hey, Rodrigo! Indeed it would. Great, subtle catch. Thanks for defending the functionality, and also for being nice about it. It hadn't struck me that one might use the individual symlinking of _virtual_includes to get bazel to enforce the header inclusion restrictions Bazel otherwise doesn't without `--features=layering_check`. Closing down this ham-handed attempt at helping out. 

Thanks again for your careful eye!
Chris

PR #934, opened 2023-07-05
The aarch64 version test case fails

cx421774557:
The aarch64 version test case fails, need to determine whether it is aarch64


sergiud:
Unfortunately, we cannot accept the PR as you did not sign the CLA.

PR #939, opened 2023-08-03
SendEmail: Protect users against vulnerable logmailers

philwo:
glog is used on a variety of systems, and we must assume that some of them still use vulnerable mailers that have bugs or "interesting features" such as https://nvd.nist.gov/vuln/detail/CVE-2004-2771.

Let's protect users against accidental shell injection by validating the email addresses against a slightly stricter version of the regex used by HTML5 to validate addresses[1].

This should prevent triggering any unexpected behavior in these tools.

Also add some basic unit tests for the SendEmail method.

[1] https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -60,6 +60,7 @@
 #include <cerrno>                   // for errno
 #include <sstream>
 #include <regex>
+#include <cctype> // for std::isspace
 #ifdef GLOG_OS_WINDOWS
 #include "windows/dirent.h"
 #else
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -2206,27 +2206,14 @@ static string ShellEscape(const string& src) {
   }
   return result;
 }
-#endif
-
-// Trim whitespace from the start of the provided string.
-static inline void ltrim(std::string &s) {
-  s.erase(s.begin(), std::find_if(s.begin(), s.end(), [](char ch) {
-      return std::isspace(ch) == 0;
-  }));
-}
-
-// Trim whitespace from the end of the provided string.
-static inline void rtrim(std::string &s) {
-  s.erase(std::find_if(s.rbegin(), s.rend(), [](char ch) {
-      return std::isspace(ch) == 0;
-  }).base(), s.end());
-}
 
 // Trim whitespace from both ends of the provided string.
 static inline void trim(std::string &s) {
-  rtrim(s);
-  ltrim(s);
+  const auto toRemove = [](char ch) { return std::isspace(ch) == 0; };
+  s.erase(s.begin(), std::find_if(s.begin(), s.end(), toRemove));
+  s.erase(std::find_if(s.rbegin(), s.rend(), toRemove).base(), s.end());
 }
+#endif
 
 // use_logging controls whether the logging functions LOG/VLOG are used
 // to log errors.  It should be set to false when the caller holds the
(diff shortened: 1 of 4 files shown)


philwo:
I'm not sure why the AppleClang CI job failed due to some issues with calculating code coverage. If someone could help me fix this, I'd appreciate it.

src/logging.cc
```diff
@@ -2207,13 +2208,72 @@ static string ShellEscape(const string& src) {
 }
 #endif
 
+// Trim whitespace from the start of the provided string.
```
ukai, line 2211: #ifndef GLOG_OS_EMSCRIPTEN 
?

https://github.com/google/glog/actions/runs/5749819819/job/15585367819?pr=939

ukai:
> I'm not sure why the AppleClang CI job failed due to some issues with calculating code coverage.

hmm. using regexp is problematic??

src/logging.cc
```diff
@@ -2207,13 +2208,72 @@ static string ShellEscape(const string& src) {
...
 // log_mutex.
 static bool SendEmailInternal(const char*dest, const char *subject,
                               const char*body, bool use_logging) {
 #ifndef GLOG_OS_EMSCRIPTEN
+  // We validate the provided email addresses using the same regular expression
+  // that HTML5 uses[1], except that we require the address to start with an
+  // alpha-numeric character. This is because we don't want to allow email
+  // addresses that start with a special character, such as a pipe or dash,
+  // which could be misunderstood as a command-line flag by certain versions
+  // of `mail` that are vulnerable to command injection.[2]
```
ukai, line 2242: for this, it might be sufficient to check the first letter is alphanum ?

philwo: Thank you! That's a very good observation! Let me think a bit about whether we should go for the minimum necessary fix here, or whether we should use the opportunity to validate the email addresses in general..

koto (approved):
LGTM

src/logging.cc
```diff
@@ -2207,13 +2208,72 @@ static string ShellEscape(const string& src) {
...
+      return std::isspace(ch) == 0;
+  }).base(), s.end());
+}
+
+// Trim whitespace from both ends of the provided string.
+static inline void trim(std::string &s) {
+  rtrim(s);
+  ltrim(s);
+}
+
```
sergiud, line 2230: `ltrim` and `rtrim` do not seem to be used anywhere except in `trim`. In light of this, the code can be greatly simplified:
```suggestion
// Trim whitespace from both ends of the provided string.
static inline void trim(std::string &s) {
  const auto toRemove = [](char ch) { return std::isspace(ch) == 0; };
  s.erase(s.begin(), std::find_if(s.begin(), s.end(), toRemove));
  s.erase(std::find_if(s.rbegin(), s.rend(), toRemove).base(), s.end());
}

```

src/logging.cc
```diff
@@ -2207,13 +2208,72 @@ static string ShellEscape(const string& src) {
...
+  // [2] e.g. https://nvd.nist.gov/vuln/detail/CVE-2004-2771
+  const std::regex pattern("^[a-zA-Z0-9]"
+      "[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]*@[a-zA-Z0-9]"
+      "(?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9]"
+      "(?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$");
+
   if (dest && *dest) {
+    // Split the comma-separated list of email addresses, validate each one and
+    // build a sanitized new comma-separated string without whitespace.
+    std::stringstream ss(dest);
```
sergiud, line 2253:
```suggestion
    std::istringstream ss(dest);
```

src/logging.cc
```diff
@@ -2207,13 +2208,72 @@ static string ShellEscape(const string& src) {
...
 #ifndef GLOG_OS_EMSCRIPTEN
+  // We validate the provided email addresses using the same regular expression
+  // that HTML5 uses[1], except that we require the address to start with an
+  // alpha-numeric character. This is because we don't want to allow email
+  // addresses that start with a special character, such as a pipe or dash,
+  // which could be misunderstood as a command-line flag by certain versions
+  // of `mail` that are vulnerable to command injection.[2]
+  // [1] https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address
+  // [2] e.g. https://nvd.nist.gov/vuln/detail/CVE-2004-2771
+  const std::regex pattern("^[a-zA-Z0-9]"
```
sergiud, line 2245: Move `pattern` into the `if` statement below since it is not used anywhere else.

sergiud (changes requested):

src/logging.cc
philwo: Thank you, that's a good idea!

philwo:
I addressed all comments, but the coverage check on macOS still fails. I don't see how my change is related to it, and if I run the lcov commands locally on my Mac with latest Xcode they fail at even more places.

Is this just broken in general at the moment due to latest Xcode being stricter / buggy?

philwo:
@sergiud Do you have some advice what to do about the code coverage errors?

sergiud:
Thanks for bringing the issue to my attention. Let me look into it.

sergiud (changes requested):
We can ignore the coverage problem as it is a separate issue.

src/logging.cc
```diff
@@ -2205,6 +2206,13 @@ static string ShellEscape(const string& src) {
   }
   return result;
 }
+
+// Trim whitespace from both ends of the provided string.
+static inline void trim(std::string &s) {
+  const auto toRemove = [](char ch) { return std::isspace(ch) == 0; };
```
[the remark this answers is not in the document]
philwo: Thank you! I added the header, let's see if it builds. :)

philwo:
Seems like we're good? You mentioned that the coverage failure on Apple is a separate issue, and the Windows clang build failure on Bazel CI also happens on the master branch, so seems unrelated. 😊

philwo:
My friends in the Bazel team said that the Windows clang failure should be fixed in Bazel 6.4.0 LTS: https://github.com/bazelbuild/bazel/pull/19430

sergiud (approved):

PR #949, opened 2023-09-07
fix(CMakeLists): build fails when using cmake on mac with Xcode as ge…

cn-nice:
When you use cmake to build with xcode as a generator under Mac, the build will fail.
See https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -495,6 +495,22 @@ target_compile_features (glog_internal PUBLIC $<TARGET_PROPERTY:glog,COMPILE_FEA
 add_library (glog
   $<TARGET_OBJECTS:glog_internal>
 )
+
+# Note: When you use cmake to build with xcode as a generator under Mac, the build will fail.
+# See https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries
+if(CMAKE_GENERATOR STREQUAL "Xcode")
+  file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeTmp)
+  set(temporary_file "${CMAKE_CURRENT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeTmp/xcode.cpp")
+
+  add_custom_command(
+    OUTPUT ${temporary_file}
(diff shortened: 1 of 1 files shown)


sergiud:
Thanks for the PR. However, you need to sign the CLA first.

It is also not clear why such a hack is necessary. Could you please explain? The CMake documentation you reference does not mention any problem with Xcode.

cn-nice:
> Thanks for the PR. However, you need to sign the CLA first.
> 
> It is also not clear why such a hack is necessary. Could you please explain? The CMake documentation you reference does not mention any problem with Xcode.

Thank you for your reply.

In the official CMake reference document, it is mentioned that: **"Some native build systems (such as Xcode) may not like targets that have only object files, so consider adding at least one real source file to any target that references $<TARGET_OBJECTS:objlib>."**. This is exactly the problem I encountered; 

When I build glog, although Xcode prompts "Build Successful", it only generates libglog_internal.a, but not libglogd.a. If you generate dynamic libraries, then it will only generate symbolic links, and the final libglogd.0.7.0.dylib is not generated.

This can be solved by adding an empty temporary file, of course there may be a better way to do this, but so far this solves my problem perfectly.

sergiud:
Thanks for the explanation. Can't we just add an existing file such as any header instead of generating a bogus one? Or better yet get rid of the glogtest target completely and use the object library target directly? In either case, I want to avoid additional logic just for Xcode support.

cn-nice:
> Thanks for the explanation. Can't we just add an existing file such as any header instead of generating a bogus one? Or better yet get rid of the glogtest target completely and use the object library target directly? In either case, I want to avoid additional logic just for Xcode support.

I understand what you mean, so I made other attempts.
Add a cpp file that already exists (header files are not allowed, because they do not constitute a compilation unit), as follows:
```cpp
add_library (glog
  $<TARGET_OBJECTS:glog_internal>
  src/logging.cc
)
```
This also solves the problem, and if you think it does, I'll resubmit the PR!

sergiud:
I pushed b8a251c091ec6ab98b43c5c70b27663b89b1c699 which deals with the issue. Please give it a try and let me know if you encounter any problems.

PR #960, opened 2023-10-06
Support truncate log file for windows

bqhuyy:
- Use `_chsize_s` (in `io.h` header) instead of `truncate` on Windows
- Reference: https://learn.microsoft.com/vi-vn/cpp/c-runtime-library/reference/chsize-s?view=msvc-150

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,6 +131,7 @@ check_function_exists (sigaltstack HAVE_SIGALTSTACK)
 check_cxx_symbol_exists (backtrace execinfo.h HAVE_EXECINFO_BACKTRACE)
 check_cxx_symbol_exists (backtrace_symbols execinfo.h
   HAVE_EXECINFO_BACKTRACE_SYMBOLS)
+check_cxx_symbol_exists (_chsize_s io.h HAVE__CHSIZE_S)
 
 # NOTE gcc does not fail if you pass a non-existent -Wno-* option as an
 # argument. However, it will happily fail if you pass the corresponding -W*
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -58,6 +58,9 @@
 #ifdef HAVE_SYSLOG_H
 # include <syslog.h>
 #endif
+#ifdef HAVE__CHSIZE_S
+#include <io.h> // for truncate log file
+#endif
 #include <vector>
 #include <cerrno>                   // for errno
 #include <sstream>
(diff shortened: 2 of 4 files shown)


src/logging.cc
```diff
@@ -2425,7 +2426,7 @@ void GetExistingTempDirectories(vector<string>* list) {
 }
 
 void TruncateLogFile(const char *path, uint64 limit, uint64 keep) {
-#ifdef HAVE_UNISTD_H
+#if defined(HAVE_UNISTD_H) || defined(GLOG_OS_WINDOWS)
```
sergiud, line 2429: I suggest integrating the detection of `_chsize_s` into `CMakeLists.txt` and using the corresponding define instead of relying on ad-hoc preprocessor logic. This way, it is easier to see whether the function is available and can be used already when configuring the project.

Something along the lines should do the trick:
```cmake
check_cxx_symbol_exists (_chsize_s io.h HAVE__CHSIZE_S)
```
For Bazel the `HAVE__CHSIZE_S` define can be added here:

https://github.com/google/glog/blob/857df01007020bc4767d9f819891fb1b2da7cb86/bazel/glog.bzl#L112-L118

sergiud (changes requested):
Thanks for the PR. Just one remark.

src/logging.cc
bqhuyy: hi @sergiud , I just added these code as your suggestion

CMakeLists.txt
```diff
@@ -131,6 +131,7 @@ check_function_exists (sigaltstack HAVE_SIGALTSTACK)
 check_cxx_symbol_exists (backtrace execinfo.h HAVE_EXECINFO_BACKTRACE)
 check_cxx_symbol_exists (backtrace_symbols execinfo.h
   HAVE_EXECINFO_BACKTRACE_SYMBOLS)
+check_cxx_symbol_exists (_chsize_s io.h HAVE__CHSIZE_S)
```
sergiud, line 134: Thanks for the update!

One thing that is still missing is a `#cmakedefine` entry for `HAVE__CHSIZE_S` in `src/config.h.cmake.in` similar to this use:

https://github.com/google/glog/blob/5d4fa3d7742ed89f975730c73250bbe1d3ab1521/src/config.h.cmake.in#L130-L131

sergiud (changes requested):

CMakeLists.txt
bqhuyy: hi @sergiud , I have already updated it

sergiud (approved):

PR #972, opened 2023-10-10
LogCleaner: Fix the scenario when FLAGS_log_dir has no '/' suffix

### LINKED ISSUE #971: LogCleaner can not delete old logs
glog version: 0.6.0

My simplified reproduction process is as follows:

1. Old logs in '/tmp/glogs'：
```
#ll -rt /tmp/glogs/
total 32776
-rw-r--r-- 1 root root 33554535 Oct  1 00:00 test-cleaner.28f8271ff892.root.log.INFO.20231001-000000.65295
```

2. Code that can reproduce the problem (simplified):

```cpp
#include <string>

#include <glog/logging.h>

static constexpr const char* kLogPrefix = "test-cleaner";
static constexpr const char* kLogDirEnv = "LOG_DIR";
const std::string kLogContent(100, 'a');

int main() {
  // Init logging
  FLAGS_log_dir = std::getenv(kLogDirEnv);
  FLAGS_max_log_size = 32;  // 32 MB

  google::InitGoogleLogging(kLogPrefix);
  google::EnableLogCleaner(3);

  for (int i = 0; i < 1'000; ++i) {
    LOG(INFO) << kLogContent;
  }

  google::DisableLogCleaner();
  google::ShutdownGoogleLogging();
}
```

3. Environment variables：LOG_DIR=/tmp/glogs

It is set above to delete files that are more than **3 days old**, but they are not deleted.

lingbin:
Fixes #971

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -2397,8 +2397,16 @@ const vector<string>& GetLoggingDirectories() {
     logging_directories_list = new vector<string>;
 
     if ( !FLAGS_log_dir.empty() ) {
-      // A dir was specified, we should use it
-      logging_directories_list->push_back(FLAGS_log_dir);
+      // A dir was specified, we should use it, and make sure to end with
+      // a directory delimiter.
+      const char* const dir_delim_end =
+        possible_dir_delim + sizeof(possible_dir_delim);
+      if (std::find(possible_dir_delim, dir_delim_end,
+            FLAGS_log_dir.back()) == dir_delim_end) {
+        logging_directories_list->push_back(FLAGS_log_dir + "/");
+      } else {
+        logging_directories_list->push_back(FLAGS_log_dir);
+      }
     } else {
       GetTempDirectories(logging_directories_list);
 #ifdef GLOG_OS_WINDOWS
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -2397,12 +2397,9 @@ const vector<string>& GetLoggingDirectories() {
     logging_directories_list = new vector<string>;
 
     if ( !FLAGS_log_dir.empty() ) {
-      // A dir was specified, we should use it, and make sure to end with
-      // a directory delimiter.
-      const char* const dir_delim_end =
-        possible_dir_delim + sizeof(possible_dir_delim);
-      if (std::find(possible_dir_delim, dir_delim_end,
-            FLAGS_log_dir.back()) == dir_delim_end) {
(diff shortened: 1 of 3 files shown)


lingbin:
@sergiud  Could you please help to take a look? Thanks.

src/logging.cc
```diff
@@ -2397,8 +2397,16 @@ const vector<string>& GetLoggingDirectories() {
...
 
     if ( !FLAGS_log_dir.empty() ) {
-      // A dir was specified, we should use it
-      logging_directories_list->push_back(FLAGS_log_dir);
+      // A dir was specified, we should use it, and make sure to end with
+      // a directory delimiter.
+      const char* const dir_delim_end =
+        possible_dir_delim + sizeof(possible_dir_delim);
+      if (std::find(possible_dir_delim, dir_delim_end,
+            FLAGS_log_dir.back()) == dir_delim_end) {
```
sergiud, line 2405:
```suggestion
      // Ensure the specified path ends with a directory delimiter
      if (std::find(std::begin(possible_dir_delim), std::end(possible_dir_delim),
            FLAGS_log_dir.back()) == std::end(possible_dir_delim)) {
```
You may need to include `<iterator>`.

sergiud (changes requested):
Thanks for the PR! One suggestion.

src/logging.cc
lingbin: Thanks for your review, already addressed comment.  

After testing `<iterator>` is not needed. My guess is that it is already included in an existing header file.
From https://en.cppreference.com/w/cpp/iterator/begin
It can be seen that `std::begin` and `std::end` are defined in many header files.

sergiud: That's not how it works. You do need `<iterator>` which defines `std::begin` and `std::end` for statically sized arrays  (see [IWYU](https://github.com/include-what-you-use/include-what-you-use#include-what-you-use) why relying on indirection is a bad style). By 'may' I meant to say that I did not check whether `<iterator>` is already included in the touched translation unit.

sergiud: Just checked `logging.cc`. `<iterator>` must definitely be included since it is not present there yet.

lingbin: Thanks for your further explanation. `<iterator>` has been added.

sergiud (approved):

sergiud:
Thanks!

PR #87, opened 2016-03-02
symbolize_unittest: make it a bit more portable

vapier:

PR #985, opened 2023-12-09
symbolize_unittest: make it a bit more portable

vapier:


vapier:
rebase of #87

sergiud:
Thanks for the PR.

Could you please rebase and push the commit again as I no longer seem to have access to your branch after force pushing.

vapier:
pretty sure you can fetch the commit you clobbered from the remote

```
$ git fetch https://github.com/vapier/glog 233df366d56df9d3d32898451f2ba1ace7e234de
$ git cherry-pick FETCH_HEAD
```

sergiud:
That's not the issue. I simply can't push to your `master` since the branch no longer contains the PR commit.

For future reference, please submit your changes from a dedicated feature branch to avoid confusion that local and remote branches with the same name can cause.

vapier:
yeah, that's not really a thing.  don't force push if you don't want to clobber/rewrite a ref.

i can rebase the original commit onto the latest branch & push to my repo again.  but i can't re-open PR's in this project.  i can only file new ones.

sergiud:
> yeah, that's not really a thing. don't force push if you don't want to clobber/rewrite a ref.

What is not a thing? Force pushing is perfectly fine and necessary when rebasing which is what I did. I simply messed the local ref when pushing.

This works know. Thanks.

vapier:
requiring people use non-default branches to generate PRs to somehow protect against bad force pushes.  you force pushed the wrong state to a remote branch.  giving it a diff name wouldn't have helped -- you would have typed the remote branch in order to rewrite it.

but whatever, let's just agree to disagree and move on.

sergiud:
That's fine. I'm eager to learn.

> giving it a diff name wouldn't have helped

Why not? Compare
```bash
git push git@github.com:vapier/glog.git master:master # whoops
```
vs
```bash
git push git@github.com:vapier/glog.git portable-labels
```
If you have a better suggestion how to avoid such a mishap, please let me know.

Issue #872, opened 2022-11-08
TID in the files and on the stack trace not matching


amandeepgautam:
In the stack trace, we see something like this:
*** SIGSEGV (@0x149d) received by PID 22298 (TID 0x36087700) from PID 5277; stack trace: ***

In the files, we see:
I1107 17:38:43.965498 22378 xx.cc:nnn] <log message>

22378 is supposed to be the thread id and so is:  0x36087700. I skimmed over the code and seems like 2 different functions are used to compute the values:

(a) https://github.com/google/glog/blob/a34226ca94be8faae363b0599e5c4cdb7767774a/src/signalhandler.cc#L211
(b) https://github.com/google/glog/blob/f4dd77ae6b8e5f8e1323c929e81dbc40ff58fc7b/src/utilities.cc#L259

What is the relation between the two? How do I know which TID caused the crash?

sergiud:
The difference is expected as per [`gettid` documentation](https://man7.org/linux/man-pages/man3/pthread_self.3.html):

> The thread ID returned by this call is not the same thing as a POSIX thread ID (i.e., the opaque value returned by [pthread_self(3)](https://man7.org/linux/man-pages/man3/pthread_self.3.html)).

The answer to your question depends on how you intend to identify your threads.

amandeepgautam:
@sergiud I want to look at which thread caused the crash. Is there a way to know that?
Out of curiosity, why not call the same function at both places?

sergiud:
> I want to look at which thread caused the crash. Is there a way to know that?

I'm not sure I understand what are you asking given you have two IDs at your disposal. Choose the one that suits your tooling. 

> Out of curiosity, why not call the same function at both places?

This is likely an inconsistency that arose over time. You are welcome to provide a fix.

amandeepgautam:
> I'm not sure I understand what are you asking given you have two IDs at your disposal. Choose the one that suits your tooling.

@sergiud We have a stack trace of the crash with an ID: 0x36087700. I want to see the log files and see what this thread was doing before crashing. In a multi-threaded env. this can be really helpful.

> This is likely an inconsistency that arose over time. You are welcome to provide a fix.

Which function is preferable: GetTID or pthread_self? I will see if I can fix it.

sergiud:
For consistency and information gain both functions should be used.

PR #1008, opened 2023-12-27
publish TID in 2 formats known formats on a crash

amandeepgautam:
Fix for: https://github.com/google/glog/issues/872

### DIFF
--- a/src/signalhandler.cc
+++ b/src/signalhandler.cc
@@ -211,6 +211,8 @@ void DumpSignalInfo(int signal_number, siginfo_t *siginfo) {
   pthread_t id = pthread_self();
   formatter.AppendUint64(
       reinterpret_cast<uint64>(reinterpret_cast<const char*>(id)), 16);
+  formatter.AppendString("/");
+  formatter.AppendUint64(static_cast<uint64>(GetTID()), 10);
   formatter.AppendString(") ");
   // Only linux has the PID of the signal sender in si_pid.
 #ifdef GLOG_OS_LINUX


src/signalhandler.cc
```diff
@@ -211,6 +211,8 @@ void DumpSignalInfo(int signal_number, siginfo_t *siginfo) {
   pthread_t id = pthread_self();
   formatter.AppendUint64(
       reinterpret_cast<uint64>(reinterpret_cast<const char*>(id)), 16);
+  formatter.AppendString("/");
+  formatter.AppendUint64(static_cast<uint64>(GetTID()), 10);
```
sergiud, line 215: I would suggest using a format similar to the backtrace. Instead of delineating the PID and TID by `/`, a `PID (TID)` formatting (TID being to base 16 with the `0x` prefix) seems to be more appropriate.
```suggestion
  formatter.AppendString(" (");
  formatter.AppendUint64(static_cast<uint64>(GetTID()), 16);
  formatter.AppendString(")");
```

sergiud (changes requested):
Thanks for the PR. I have one remark.

Please also update the commit message to mention the changes to be POSIX threads specific and the two thread identifiers being POSIX thread ID and kernel thread ID.

src/signalhandler.cc
amandeepgautam: Doing it in base 16 may not be ideal as glog internally publishes base 10 POSIX thread IDs. For example, a line would be:

I1231 01:02:58.835251  1681 kkkk.cc:1591] kkkkkk

So, if I use 0x format, one would have to convert it to 1681 (which would be one more step).

amandeepgautam: Additionally, in the suggested format we would have 2 nested parenthesis like:
(TID 0x7f7666a8b980 (0x2f424...))

I find 2 nested brackets slightly odd. But am okay with what you say.

amandeepgautam: Or we can modify the comment to be:
received by PID 2216 in kernel thread ID 0x7f7666a8b980 (POSXI ID: <...>) from PID 2216

from the current version:
received by PID 2216 (TID 0x7f7666a8b980) from PID 2216

sergiud: I only referred to TID to be base 16 which is the format currently used by glog. The base 10 format of POSIX thread ID should remain. It is important that the format is consistent throughout with respect to radix, ID order, possible labels, etc.

Nested parentheses are of course best avoided. You can merge the identifiers and put them into a single pair of round brackets joined by `, ` (comma and a single space).

amandeepgautam: > I only referred to TID to be base 16 which is the format currently used by glog. The base 10 format of POSIX thread ID should remain.

I am slightly confused now. Please bear with me as I am not familiar with the codebase. As per [this line](https://github.com/google/glog/blob/d29d02e8afc8afc87726c47dd91a89a5a51c84ec/src/logging.cc#L1674), GetTID seems to be the one returning unsigned int. So static_cast<uint64>(GetTID()), 10); which is my original implementation seems reasonable. In the suggestion in the comment, we have static_cast<uint64>(GetTID()), 16) which seems to be suggesting that I should convert it to base16, which is not my intention. Can you please let me know what am I missing?


> Nested parentheses are of course best avoided. You can merge the identifiers and put them into a single pair of round brackets joined by , (comma and a single space).

Ack.

sergiud: Yes, you are correct. I looked through the code and the TID formatting is indeed inconsistent. I therefore suggest keeping the existing format variants and use base 10 representation for the kernel thread IDs consistently instead.

Please note that #1019 removed the use of POSIX threads. There are now conflicts in your branch which you should rebase and resolve.

sergiud:
Closing as abandoned. You can rebase anytime if you want to continue working on this PR.

PR #1067, opened 2024-01-22
docs: give a example to show how to import glog by add_subdirectory

valaxkong:
Change-Id: I2e77e155c134d36f36ff38f3d3bc396a48b07ff6

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -41,7 +41,7 @@ jobs:
 
       - name: Cache GTest
         id: cache-gtest
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: gtest/
           key: ${{runner.os}}-gtest-1.11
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -79,6 +79,27 @@ Alternatively, glog can be incorporated into using the CMake command
 project by replacing the :cmake:`find_package` call from the previous snippet by
 :cmake:`add_subdirectory`. The :cmake:`glog::glog` target is in this case an
 :cmake:`ALIAS` library target for the ``glog`` library target.
+Here is an example which using :cmake:`FetchContent` and :cmake:`add_subdirectory`
```
sergiud, line 82: Do we really need to show a complete example for how to use `FetchContent`? This is essentially an exact mirror of the example from [CMake reference documentation](https://cmake.org/cmake/help/latest/module/FetchContent.html#overview). In that sense, a single sentence with a link to `FetchContent` example should be sufficient.

sergiud (changes requested):
Thanks for the PR.

I do not believe we should mirror the CMake documentation. An example of using `FetchContent` is better suited for the Wiki section of glog.

valaxkong:
@sergiud 
New PR: https://github.com/google/glog/pull/1068

PR #1069, opened 2024-01-27
feat: supports unicode paths

xfc1939:
solved the questions about  #998

### DIFF
--- a/src/base/commandlineflags.h
+++ b/src/base/commandlineflags.h
@@ -108,12 +108,12 @@
     }                                                           \
     using fLS::FLAGS_##name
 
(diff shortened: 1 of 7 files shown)


xfc1939:
I used a marco that named GLOG_DEFINE_wstring in flags.cc,  this marco defined in commandlineflags.h. 
It‘s ok on CMake, but it's failed on bazel. because it's used gflags/gflags.h on bazel. 
Have anyone give me a solution?

sergiud:
Please do not create multiple PRs for the same issue because the comments are lost.

Let's continue in #998 where we left off.

PR #1083, opened 2024-03-11
OverrideLogCleanerFilePathPrefix to override the files the log cleane…

mturnock:
…r removes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -728,6 +728,11 @@ if (BUILD_TESTING)
 
   target_link_libraries (cleanup_with_relative_prefix_unittest PRIVATE glog_test)
 
+  add_executable (cleanup_with_file_path_override_unittest
+    src/cleanup_with_file_path_override_unittest.cc)
+
+  target_link_libraries (cleanup_with_file_path_override_unittest PRIVATE glog_test)
+
   set (CLEANUP_LOG_DIR ${glog_BINARY_DIR}/cleanup_tests)
 
   add_test (NAME cleanup_init COMMAND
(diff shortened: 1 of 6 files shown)


mturnock:
My requirement was that I delete old log files, regardless of log level.
I was looking to achieve something similar to https://github.com/google/glog/pull/1062, but I don't think that approach is very easy.  So instead I extended the interface to override the files the log cleaner cleans, giving flexibility to users of the log cleaner on what they want cleaning.

mturnock:
> Thanks for the PR!
> 
> However, I'm not sure I understand why we need to extended the public API instead of fixing the underlying issue?

I don't think I have come up with a better alternative.  I'm open to suggestions.

Some possible thoughts:
1. Make next_cleanup_time_ a map, 1 entry per base_filename.  Pro: no change to interface.  Con: only deletes old log of a particular level when a new log line of that level is made.
2. Add a registration system to log cleaner from each LogDestination.  Con: seemed a bit gnarly.

mturnock:
The suggestion from https://github.com/google/glog/pull/1062 is that we can fix IsLogFromCurrentProject, but I don't know if that is possible in general because the user can set any log name style they want with SetLogDestination

sergiud:
I would prefer no additional methods if possible since using log cleaner will become less intuitive than it already is.

While the log destination can be different the suffix is always the same. Combined with the observation that log severities are known at compile time determining candidates for deletion should be doable.

mturnock:
Consider:

  SetLogDestination(GLOG_INFO, "/path/to/logs/applog-carrots-");
  SetLogDestination(GLOG_WARNING, "/path/to/logs/applog-peas-");
  SetLogDestination(GLOG_ERROR, "/path/to/logs/applog-corn-");
  SetLogFilenameExtension(".log");
  EnableLogCleaner(std::chrono::days(7));

If we had a registration system, then that would work, but trying to fix IsLogFromCurrentProject seems futile.

sergiud:
As I see it, an internal registration system is still a better solution than shifting the burden of managing log file paths onto users.

mturnock:
OK, thanks.  I've closed this one, and opened a new one to do the registration system approach. https://github.com/google/glog/pull/1086

PR #1087, opened 2024-03-13
Check if __argv is null

mturnock:
__argv can be null when running unicode programs compiled with msvc

### DIFF
--- a/src/utilities.cc
+++ b/src/utilities.cc
@@ -237,7 +237,11 @@ const char* ProgramInvocationShortName() {
 #elif defined(HAVE___PROGNAME)
   return __progname;
 #elif defined(HAVE___ARGV)
-  return const_basename(__argv[0]);
+  if (__argv) {
+    return const_basename(__argv[0]);
+  } else {
+    return "UNKNOWN";
+  }
 #else
   return "UNKNOWN";
 #endif


src/utilities.cc
```diff
@@ -237,7 +237,11 @@ const char* ProgramInvocationShortName() {
 #elif defined(HAVE___PROGNAME)
   return __progname;
 #elif defined(HAVE___ARGV)
-  return const_basename(__argv[0]);
+  if (__argv) {
+    return const_basename(__argv[0]);
+  } else {
+    return "UNKNOWN";
+  }
```
sergiud, line 244:
```suggestion
  if (__argv != nullptr) {
    return const_basename(__argv[0]);
  }
```
Please avoid copy-and-pasting the return branch below.

With this change, unicode applications will use `UNKNOWN` as the program name, correct? That's not really appealing. Shouldn't glog fallback to `__wargv[0]` converted to the current code page instead?

sergiud (changes requested):
Thanks for the PR!

In my opinion, the proposed workaround is of limited use. The actual fix should be symmetrical to the non-unicode case.

src/utilities.cc
mturnock: Thanks.  I'm not sure that's quite right (what should the fn return without else?; the #else is preprocessor), but i agree, the duplication is unfortunate.  Perhaps you can suggest an alternative?  I think unicode programs not crashing is more appealing than implementing more features, but if that code is already in the repo to do the conversion, let me know.

sergiud: The suggestion is not complete. You also need to adjust the preprocessor directives to obtain duplication free code structure.

>  I think unicode programs not crashing is more appealing than implementing more features, but if that code is already in the repo to do the conversion, let me know.

Let me put it differently: a half baked change is not something I would consider here. This needs a proper fix.

There are two options here:
1. Use `WideCharToMultiByte` to convert `__wargv[0]` to user's code page
2. Use `GetModuleFileNameA` directly instead of converting `__wargv`

sergiud:
Closing as abandoned. Please let me know if you want to continue working on this PR.

Issue #858, opened 2022-09-07
Comment describing SetVLOGLevel is not accurate


hari90:
Currently the comment states the following:
// NOTE: To change the log level for VLOG(_IS_ON) sites
//	 that have already executed after/during InitGoogleLogging,
//	 one needs to supply the exact --vmodule pattern that applied to them.
//       (If no --vmodule pattern applied to them
//       the value of FLAGS_v will continue to control them.)

But https://github.com/google/glog/pull/650 added support for updating vmodule levels after vmodule level has been cached.

The comment should be
// NOTE: To change the log level for VLOG(_IS_ON) sites
//	 that matched a prior --vmodule pattern then the exact pattern has to be supplied again.
//       (If no --vmodule pattern applied to them before then the first matching pattern will start to control them.)

sergiud:
@romange Could you please verify the suggestion?

hari90:
@sergiud @romange I submitted https://github.com/google/glog/issues/1109 for this. Can you please review it?

PR #1109, opened 2024-06-21
Fix comment NOTE about SetVLOGLevel

hari90:
Fix the comment about `SetVLOGLevel` according to https://github.com/google/glog/pull/650 which added support for updating vmodule levels after vmodule level has been cached.

Fixes https://github.com/google/glog/issues/858

### DIFF
--- a/src/glog/vlog_is_on.h
+++ b/src/glog/vlog_is_on.h
@@ -103,11 +103,10 @@ namespace google {
 // Set VLOG(_IS_ON) level for module_pattern to log_level.
 // This lets us dynamically control what is normally set by the --vmodule flag.
 // Returns the level that previously applied to module_pattern.
-// NOTE: To change the log level for VLOG(_IS_ON) sites
-//	 that have already executed after/during InitGoogleLogging,
-//	 one needs to supply the exact --vmodule pattern that applied to them.
-//       (If no --vmodule pattern applied to them
-//       the value of FLAGS_v will continue to control them.)
+// NOTE: To change the log level for VLOG(_IS_ON) sites that matched a prior
+// --vmodule pattern then the exact pattern has to be supplied again. (If no
+// --vmodule pattern applied to them before then the first matching pattern will
+// start to control them.)
 extern GLOG_EXPORT int SetVLOGLevel(const char* module_pattern, int log_level);
 
 // Various declarations needed for VLOG_IS_ON above: =========================


src/glog/vlog_is_on.h
```diff
@@ -103,11 +103,10 @@ namespace google {
 // Set VLOG(_IS_ON) level for module_pattern to log_level.
 // This lets us dynamically control what is normally set by the --vmodule flag.
 // Returns the level that previously applied to module_pattern.
-// NOTE: To change the log level for VLOG(_IS_ON) sites
-//	 that have already executed after/during InitGoogleLogging,
-//	 one needs to supply the exact --vmodule pattern that applied to them.
-//       (If no --vmodule pattern applied to them
-//       the value of FLAGS_v will continue to control them.)
+// NOTE: To change the log level for VLOG(_IS_ON) sites that matched a prior
+// --vmodule pattern then the exact pattern has to be supplied again. (If no
```
sergiud, line 107: I do not understand this part. Is the `then` part redundant?

sergiud (changes requested):
Thanks for the PR.

Doesn't it make more sense to supply the updated information in the user guide, for instance, by updating [this](https://github.com/google/glog/blob/master/docs/logging.md#verbose-logging) section?

hari90:
> Thanks for the PR.
> 
> Doesn't it make more sense to supply the updated information in the user guide, for instance, by updating [this](https://github.com/google/glog/blob/master/docs/logging.md#verbose-logging) section?

I have put a basic description of `SetVLOGLevel` in the doc and more detailed information in the .h file. Let me know if this looks good enough.

docs/logging.md
```diff
@@ -247,9 +247,11 @@ The wildcarding functionality 3. supports both `*` (matches 0 or more
 characters) and `?` (matches any single character) wildcards. Please also refer
 to [command line flags](flags.md) for more information.
 
+You can dynamically control the level of a module using the `#!cpp SetVLOGLevel` function.
+
 There's also `#!cpp VLOG_IS_ON(n)` "verbose level" condition macro. This macro
-returns `#!cpp true` when the `--v` is equal to or greater than `n`. The macro can be
-used as follows:
+returns `#!cpp true` when the matching `--vmodule`, else `--v` is equal to or greater than `n`.
```
sergiud, line 253: This description seems to be broken now. The `else` part looks out of place. What am I missing?

sergiud (changes requested):
Thank you for updating the docs. However, there seems to some kind of editing error.

docs/logging.md
hari90: I was trying to convey the fact that if there is no matching --vmodule then --v is used. Should we just change it to "when the level is equal to or greater than"?

hari90: ping on this

hari90: @sergiud ping on this

PR #1110, opened 2024-06-22
Restore "Check failure stack trace" message on LOG(FATAL)

AustinSchuh:
https://github.com/google/glog/pull/1074 refactored some of the code to enable the failure function to throw.  This made it so when the LogMessageFatal class was used, the error ended up being printed differently.

Before:
  LOG_AT_LEVEL(google::LogSeverity::FATAL) << "Crash: Hello world!"; ->

  F20240621 18:12:44.710584 139620827212672 log_demo.cc:16] Crash: Hello world!
  *** Check failure stack trace: ***
      @     0x559e2704711a
      @     0x7efc01fac24a
      @     0x7efc01fac305
      @     0x559e27046dd5
  Aborted

  LOG(FATAL) << "Crash: Hello world!"; ->

  F20240621 18:13:05.760556 140518290856832 log_demo.cc:16] Crash: Hello world!
      @     0x55cdc2475130
      @     0x7fccf6fb324a
      @     0x7fccf6fb3305
      @     0x55cdc2474df5
  Aborted

With this patch, they both produce the same output.


src/logging_unittest.cc
```diff
@@ -1586,3 +1586,12 @@ TEST(Logging, FatalThrow) {
   ScopedExit<decltype(restore_fail)> restore{restore_fail};
   EXPECT_THROW({ LOG(FATAL) << "must throw to fail"; }, std::logic_error);
 }
+
+TEST(DeathLogging, ErrorMessage) {
+  ASSERT_DEATH({ LOG(FATAL) << "foo"; }, "Check failure stack trace");
+  ASSERT_DEATH(
+      {
+        LOG_AT_LEVEL(google; : LogSeverity::FATAL) << "foo";
```
sergiud, line 1594: This is not a valid C++ syntax which is probably the reason for the failing CI. I believe you wanted something like this:
```suggestion
        LOG_AT_LEVEL(google::LogSeverity::FATAL) << "foo";
```

sergiud (changes requested):
Thank you for the PR.

Unfortunately, the CI builds are now failing. Please inspect the runner output to identify the reason.

src/logging_unittest.cc
AustinSchuh: Whops, done.  I was having trouble running tests locally with bazel.

Issue #1111, opened 2024-07-05
Stacktrace is not printed by failure signal handler when built with Bazel


kuntryn:
When glog is built with Bazel, failure signal handler is unable to print stack trace.

*** Aborted at 1720166741 (unix time) try "date -d @1720166741" if you are using GNU date ***
PC: @                0x0 (unknown)
*** SIGSEGV (@0x3e8000070c7) received by PID 28701 (TID 0x7ffff478aa40) from PID 28871; stack trace: ***

This problem has occurred before and has been fixed, but now it has reappeared. https://github.com/google/glog/issues/346

sergiud:
According to #851 this should be working. What glog version are using?

kuntryn:
My mistake. The stacktrace is indeed printed, but it's not interpreted as symbols (function names)，`LOG(FATAL)` doesn't interpret either.

I'm using version 0.7.1 on bazel central registry: https://registry.bazel.build/modules/glog, building in debug mode `bazel build @//:target -c dbg`

```
*** Aborted at 1720403914 (unix time) try "date -d @1720403914" if you are using GNU date ***
PC: @                0x0 (unknown)
*** SIGSEGV (@0x3e800000c85) received by PID 3038 (TID 0x7ffff478aa40) from PID 3205; stack trace: ***
    @     0x5555571b5835 (unknown)
    @     0x5555571b57ed (unknown)
    @     0x5555571b57b4 (unknown)
    @     0x5555571b5784 (unknown)
    @     0x5555571b5751 (unknown)
    @     0x7ffff70a3ee8 (unknown)
    @     0x5555571b4c3b (unknown)
    @     0x5555571b5669 (unknown)
    @     0x5555571b490a (unknown)
    @     0x7ffff704c520 (unknown)
    @     0x7ffff712fe2e (unknown)
    @     0x555557277471 (unknown)
    @     0x55555727b48a (unknown)
    @     0x555557271bc7 (unknown)
    @     0x555557271a8e (unknown)
    @     0x5555563c94e1 (unknown)
    @     0x5555563c9b2a (unknown)
    @     0x7ffff7033d90 (unknown)
    @     0x7ffff7033e40 (unknown)
    @     0x5555563c7de5 (unknown)
    @                0x0 (unknown)
```

PR #1116, opened 2024-07-31
feat(symbolize): Bazel build system define HAVE_SYMBOLIZE

hcoona:
`link.h` is usually provided by glibc. Current implementation of symbolize can work on `link.h`. Enhance the symbolize detection logic to support `link.h` by defining HAVE_SYMBOLIZE in bazel build description file.

The choice of altering bazel build file is for the following reasons

1. "glog moved away from using preprocessor logic in recent releases because this approach is error prone. Contributions are expected to follow the methodology for consistency reasons." according to @sergiud 
2. CMake build file have supported it.

This PR close https://github.com/google/glog/issues/1084, supersede https://github.com/google/glog/pull/1085 and https://github.com/google/glog/pull/1114. It also fixes https://github.com/google/glog/issues/1111.

### DIFF
--- a/src/symbolize.h
+++ b/src/symbolize.h
@@ -77,6 +77,10 @@
 #  error "symbolize.h" was not included correctly.
 #endif
 
+// We prefer to let the build system detect the availability of certain features
+// such as symbolization support. HAVE_SYMBOLIZE should therefore be defined by
+// the build system in general unless there is a good reason to perform the
+// detection using the preprocessor.
 #ifndef GLOG_NO_SYMBOLIZE_DETECTION
 #  ifndef HAVE_SYMBOLIZE
 // defined by gcc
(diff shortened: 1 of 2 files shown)


drigz (commented):
LGTM from a Bazel point-of-view, but leaving approval to @sergiud as he's mentioned in the comment.

src/symbolize.h
```diff
@@ -77,6 +77,8 @@
 #  error "symbolize.h" was not included correctly.
 #endif
 
+// INFO(sergiud): Add new features to build system instead of here.
+// See GitHub pull request !1114 for a detail discussion.
```
sergiud, line 81: To avoid any confusion we should entirely document the current approach:
```suggestion
// We prefer to let the build system detect the availability of certain features such as symbolization support.
// HAVE_SYMBOLIZE should therefore be defined by the build system in general unless there is a good reason to
// perform the detection using the preprocessor.
```

sergiud (changes requested):
Thanks for the PR. Bazel part is looking good. I have just a remark regarding the comment: there is no strict rule for how features should be detected. In this specific instance, however, we prefer the use of the build system for consistency and maintainability reasons.

drigz (approved):

sergiud (approved):

PR #1120, opened 2024-08-02
chore: add MODULE.bazel.lock to .gitignore

4kangjc:


drigz:
How have you tested this?

I think https://github.com/google/glog/pull/1116 might be the better fix here - but if you've tested with wasm then this is the right place to change it.

4kangjc:
> How have you tested this?
> 
> I think #1116 might be the better fix here - but if you've tested with wasm then this is the right place to change it.

I think `DHAVE_UNWIND` and `DHAVE_SYMBOLIZE` are better together, they are stacktrace related macros, maybe it would be better to move them to linux_or_darwin_copts.

4kangjc:
> How have you tested this?
> 
> I think #1116 might be the better fix here - but if you've tested with wasm then this is the right place to change it.

I just saw #1116. #1116 can indeed solve #1111

drigz:
> I think DHAVE_UNWIND and DHAVE_SYMBOLIZE are better together, they are stacktrace related macros, maybe it would be better to move them to linux_or_darwin_copts.

I have no wasm dev environment and no one has contributed a CI test, so I'd lean towards only changing the options we can test.

4kangjc:
> > I think DHAVE_UNWIND and DHAVE_SYMBOLIZE are better together, they are stacktrace related macros, maybe it would be better to move them to linux_or_darwin_copts.
> 
> I have no wasm dev environment and no one has contributed a CI test, so I'd lean towards only changing the options we can test.

OK. This PR is changed to only add MODULE.bazel.lock to .gitignore

drigz (approved):
Thank you!

.gitignore
```diff
@@ -2,3 +2,5 @@
 /build*/
 /site/
 bazel-*
+# Bzlmod lockfile
+MODULE.bazel.lock
```
sergiud, line 6: I would guess that the file can only be present in the project root only.
```suggestion
/MODULE.bazel.lock
```

sergiud (changes requested):
Thanks! I have a minor remark though.

.gitignore
4kangjc: done

sergiud (approved):
Thanks!

PR #1137, opened 2024-10-25
Fix null terminator in Demangle function.

martonka:
If the demangled name is longer than out_size, the null terminator is missing from the output. This will cause a crash in the DemangleInplace() function (symbolize.cc) when calling strlen on the buffer.

Reproduction Steps:

Use a function with a demangled name longer than 256 characters.
Attempt to log a fatal error with a stack trace.
This will cause a crash.
Example input for DemangleInplace that makes it crash (on Ubuntu):
"_ZN4kudu9FsManager9AddTenantERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEES8_St8optionalIS6_ESA_SA_"


martonka:
I added a second commit extending the SymbolizeWithDemangling (just demangle a longer function). It crashes without my fix.

src/demangle.cc
```diff
@@ -1351,6 +1351,9 @@ bool Demangle(const char* mangled, char* out, size_t out_size) {
   }
 
   std::copy_n(unmangled.get(), std::min(n, out_size), out);
```
sergiud, line 1353: I suggest storing the `std::min` result in a variable and writing the null character unconditionally.

src/symbolize_unittest.cc
```diff
@@ -150,6 +181,9 @@ TEST(Symbolize, SymbolizeWithDemangling) {
 #      if !defined(_MSC_VER) || !defined(NDEBUG)
 #        if defined(HAVE___CXA_DEMANGLE)
   EXPECT_STREQ("Foo::func(int)", TrySymbolize((void*)(&Foo::func)));
+  // Very long functions can be truncated, but we should not crash or return null
+  // Also the result should start properly.
+  EXPECT_TRUE(0 == std::string( TrySymbolize( (void*)(&Foo::longParamFunc))).find("Foo::longParamFunc("));
```
sergiud, line 186: Constructing a temporary `std::string` is unnecessary. Use `std::strstr` instead.

src/symbolize_unittest.cc
```diff
@@ -142,6 +157,22 @@ void ATTRIBUTE_NOINLINE Foo::func(int x) {
...
+		std::map<std::map<std::string, std::string>,std::map<std::string, std::string> > p2,
+		std::map<std::map<std::string, std::string>,std::map<std::string, std::string> > p3,
+		std::map<std::map<std::string, std::string>,std::map<std::string, std::string> > p4,
+		std::map<std::map<std::string, std::string>,std::map<std::string, std::string> > p5,
+		std::map<std::map<std::string, std::string>,std::map<std::string, std::string> > p6,
+		std::map<std::map<std::string, std::string>,std::map<std::string, std::string> > p7,
+		std::map<std::map<std::string, std::string>,std::map<std::string, std::string> > p8,
+		std::map<std::map<std::string, std::string>,std::map<std::string, std::string> > p9
+		) {
+  volatile auto a = p0.size() + p1.size() + p2.size() + p3.size() + p4.size() + p5.size() + p6.size() + p7.size() + p8.size() + p9.size();
```
sergiud, line 172:
```suggestion
  return p0.size() + p1.size() + p2.size() + p3.size() + p4.size() + p5.size() + p6.size() + p7.size() + p8.size() + p9.size();
```
to avoid `volatile`.

sergiud (changes requested):
Thanks for the PR! I have a couple of remarks.

src/symbolize_unittest.cc
martonka: ok

src/demangle.cc
martonka: ok

src/demangle.cc
```diff
@@ -1350,7 +1350,9 @@ bool Demangle(const char* mangled, char* out, size_t out_size) {
     return false;
   }
 
-  std::copy_n(unmangled.get(), std::min(n, out_size), out);
+  size_t copy_size = std::min(n, out_size);
+  std::copy_n(unmangled.get(), copy_size, out);
+  out[copy_size - 1] = '\0'; // Ensure terminating null if n > out_size
```
sergiud, line 1355: `copy_size - 1` will underflow if `n` is 0, no? Also `out_size` might be 0.

Possible solution:
```suggestion
  if (out_size > 0) {
    std::size_t copy_size = std::min(n, out_size - 1);
    std::copy_n(unmangled.get(), copy_size, out);
    out[copy_size] = '\0'; // Ensure terminating null if n > out_size
  }
```

sergiud (changes requested):

src/demangle.cc
martonka: n can't be 0, because __cxa_demangle() will either write a valid 0 terminated string, so n is at least 1, or nullptr is returned.

In case out_size = 0:
+ For me the windows path seems to crash too if out_size == 0 (just looking at it, I did not actually run it). 
+ If we want to handle the out_size == 0 case, then I think an "else {return false; }" is also needed for your suggestion (since no valid string is writen into out).

I think it would be simpler to just add an early return to the begining of the function ( if (out_size == 0) return false; (also fixind the windows path).

martonka: Ok. I thought it over, and neither returning true nor false is perfect. So I will just accept your changes.
Also, the Windows path is fine for out_size = 0; I misinterpreted that one.

sergiud:
I had another look at the existing unit tests and now I'm not convinced that trimming the output is a good idea. This will possibly result in some sort of ambiguity or even will confuse users.

The easiest way of fixing the issue is to simply return `false` whenever the buffer size is too small to be consistent with the behavior of the existing demangler. In the long term, however, we should switch to dynamic memory (re)allocation (e.g., by using `std::string` as the return type) to avoid insufficient buffer size situations.

src/demangle.cc
```diff
@@ -1350,7 +1351,15 @@ bool Demangle(const char* mangled, char* out, size_t out_size) {
     return false;
   }
 
-  std::copy_n(unmangled.get(), std::min(n, out_size), out);
+  if (out_size > 0) {
```
sergiud, line 1354: Return `false` early if `n > out_size`.

src/demangle.cc
```diff
@@ -1350,7 +1351,15 @@ bool Demangle(const char* mangled, char* out, size_t out_size) {
...
+    assert(n > 1);
+    // n is the size of the allocated buffer, not the length of the string.
+    // Therefore, it includes the terminating zero (and possibly additional
+    // space).
+    std::size_t copy_size = std::min(n, out_size) - 1;
```
sergiud, line 1359: All the checks are not necessary after all.

sergiud (changes requested):

sergiud:
@martonka I pushed some changes to your branch. Make sure to pull these before extending the PR.

PR #1138, opened 2024-11-05
Fix symbolize detection on macOS

take-cheeze:
Since by detecting as `UNIX` first will end up as Linux: https://cmake.org/cmake/help/latest/variable/UNIX.html


sergiud:
Thanks!

Issue #1089, opened 2024-03-19
Apply max_log_file constraint when timestamp_in_log_file=false


andrembcosta:
Currently, when the flag timestamp_in_log_file=false and max_log_file is set, instead of quitting logging or overwriting to preserve the size limit, glog just appends to the log file and eventually breaks the size limit. 

Am I missing another flag to get the intended behavior of having a single log file of limited size? Or is this an actual limitation?

This appears to be similar to https://github.com/google/glog/issues/809

HNOONa-0:
Hello @drigz,

I’ve investigated this bug and understand why it occurs. I would like to work on fixing this issue
But let's assume that I restrict each file to 1 MiB and I have exceeded that limit, I have the following options:

- Force `timestamp_in_log_file` to true. Note this can lead to duplication errors when file size constraint is too small (1 MiB)
- Add an auto-incrementing index to `LogFileObject` and start writing to a new file named `base_filename_.<index>`
- Truncate the current file once it exceeds the limit and start writing from the beginning

I assume that approach 3 works because if the original intention of the user was to keep multiple files he would use a timestamp.

drigz:
@HNOONa-0 I only have context on the Bazel integration so maybe you want @sergiud to look at that comment.

PR #1143, opened 2024-12-16
truncate single file upon creation

HNOONa-0:
A small change that writes logs to beginning of the file when its size exceeds `max_log_size` constraint.

Fixes: #1089

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -995,7 +995,22 @@ bool LogFileObject::CreateLogfile(const string& time_pid_string) {
   if (FLAGS_timestamp_in_logfile_name) {
     // demand that the file is unique for our timestamp (fail if it exists).
     flags = flags | O_EXCL;
+  } else {
+    // logs are written to a single file, where: a log file is created for the
+    // the first time or a file is being recreated due to exceeding max size
+
+    struct stat statbuf;
+    if (stat(filename, &statbuf) == 0) {
+      // truncate the file if it exceeds the max size
+      if ((static_cast<uint32>(statbuf.st_size) >> 20U) >= MaxLogSize()) {
+        flags |= O_TRUNC;
+      }
+
+      // update file length to sync file size
+      file_length_ = static_cast<uint32>(statbuf.st_size);
+    }
   }
+
   FileDescriptor fd{
       open(filename, flags, static_cast<mode_t>(FLAGS_logfile_mode))};
   if (!fd) return false;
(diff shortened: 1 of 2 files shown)


HNOONa-0:
@sergiud PTAL!

sergiud:

I'm not sure I understand the change. The log output file is truncated unconditionally; no check against `max_log_size` is performed. Can you elaborate why this still fixes the issue?

src/logging.cc
```diff
@@ -996,6 +996,11 @@ bool LogFileObject::CreateLogfile(const string& time_pid_string) {
     // demand that the file is unique for our timestamp (fail if it exists).
     flags = flags | O_EXCL;
   }
+  else{
+    // logs are written to a single file, truncate it if it exists.
+    flags = flags | O_TRUNC;
```
sergiud, line 1001: It would be nice to have a unit test for this change to avoid regressions.

sergiud (changes requested):

HNOONa-0:
> Thanks for the PR and apologies for the late reply.
> 
> I'm not sure I understand the change. The log output file is truncated unconditionally; no check against `max_log_size` is performed. Can you elaborate why this still fixes the issue?

There are two scenarios:
If `FLAGS_timestamp_in_logfile_name` is true, then a new file with a timestamp is created.

Else, it's implied that we are writing to a single log file, and in this case, whenever we are attempting to create a new file, it's because:
1. The log file is created for the first time.
2. The log file is being recreated due to length overflow in file size in this part of implementation:
https://github.com/google/glog/blob/6c5c692c8e423f651c74de9477ff0b5a59008bcc/src/logging.cc#L1110-L1117

In both these scenarios, the file should be truncated to reset the file size, hence the `O_TRUNC` flag. Is my understanding correct so far?

sergiud:
I see, thanks. Could you please extend the comment to mention that the log is recreated in the `else` branch due to size overflow as per 2?

HNOONa-0:
Hey @sergiud!
I’ve extended the comments as you suggested and added a new test method to validate these changes. In this test, I turned off the `timestamp` flag and constrained the `log_size` to 1MB. Then, I logged `2e4` lines. Although the total lines exceed the `1MB` threshold, the test ensures that the file size remains within the `1MB` limit.

HNOONa-0:
Hey @sergiud, sorry for the delay.
I've made a few changes to allow appending when the file isn't at max size, along with some minor refactoring.
The tests should pass hopefully.

sergiud:
Thanks!

Issue #882, opened 2022-12-16
Set different overduedays for different levels of logs


IoCing:
At present, we can use `EnableLogCleaner` to enable logclean and set overduedays.

but the overduedays of all level log is the same.

I think we can add some variables and methods  in logclean to support to set different overduedays for different levels of logs.

we can also support to enable or disable each level individually.

PR #1144, opened 2024-12-16
add map between severity and overdue minutes

HNOONa-0:
I have added an `unordered_map` that can store a severity and it's corresponding overdue time to the `LogCleaner` object.  The modifications should be backwards compatible. If this idea is correct, we can preform some optimizations, like keeping a cache of filepaths whose severities have been computed.

Fixes: #882

### DIFF
--- a/src/cleanup_deferred_unittest.cc
+++ b/src/cleanup_deferred_unittest.cc
@@ -1,0 +1,100 @@
+// Copyright (c) 2024, Google Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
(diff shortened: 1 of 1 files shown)


src/logging.cc
```diff
@@ -337,6 +338,16 @@ const char* GetLogSeverityName(LogSeverity severity) {
   return LogSeverityNames[severity];
 }
 
+LogSeverity FindFilepathLogSeverity(const std::string& filepath) {
+  for (int i = GLOG_INFO; i < NUM_SEVERITIES; i++) {
+    if (filepath.find(GetLogSeverityName(static_cast<LogSeverity>(i))) != std::string::npos) {
```
sergiud, line 343:
```suggestion
    if (filepath.rfind(GetLogSeverityName(static_cast<LogSeverity>(i))) != std::string::npos) {
```
Search from right since the severity is used as a suffix.

src/logging.cc
```diff
@@ -337,6 +338,16 @@ const char* GetLogSeverityName(LogSeverity severity) {
...
+      return static_cast<LogSeverity>(i);
+    }
+  }
+  // assume that the file has the INFO severity
```
sergiud, line 347: Do not assume anything and simply return a not found sentinel.

src/logging.cc
```diff
@@ -1286,7 +1299,14 @@ LogCleaner::LogCleaner() = default;
 
 void LogCleaner::Enable(const std::chrono::minutes& overdue) {
   enabled_ = true;
-  overdue_ = overdue;
+  for (int i = GLOG_INFO; i < NUM_SEVERITIES; i++) {
+    overdue_per_severity_[static_cast<LogSeverity>(i)] = overdue;
```
sergiud, line 1303: Fill the map only with the severity set by the user.

src/logging.cc
```diff
@@ -2627,6 +2654,10 @@ void EnableLogCleaner(const std::chrono::minutes& overdue) {
   log_cleaner.Enable(overdue);
 }
 
+void EnableLogCleaner(std::unordered_map<LogSeverity, std::chrono::minutes>& overdue_per_severity) {
```
sergiud, line 2657: Exposing the map as a parameter makes the calls verbose and the API usage unnecessarily complicated. Instead use the following signature:
```suggestion
void EnableLogCleaner(LogSeverity severity, const std::chrono::minutes& overdue) {
```

src/logging.cc
```diff
@@ -2627,6 +2654,10 @@ void EnableLogCleaner(const std::chrono::minutes& overdue) {
...
+  log_cleaner.Enable(overdue_per_severity);
+}
+
 void DisableLogCleaner() { log_cleaner.Disable(); }
```
sergiud, line 2661: For symmetry, there should be an API call for disabling the log cleaner for specific severity.

sergiud (changes requested):

The changes are going into the right direction. Please add some unit tests. For this, you can extend the existing log cleaner tests.

HNOONa-0:
Hey @sergiud , I have made some improvements, I switched from a map to a vector for better lookup speed and cleaner implementation. I also addressed all the issues you pointed out. PTAL!

sergiud (changes requested):

src/logging.cc
```diff
@@ -459,9 +471,7 @@ class LogCleaner {
       const string& filepath,
       const std::chrono::system_clock::time_point& current_time) const;
 
-  bool enabled_{false};
-  std::chrono::minutes overdue_{
-      std::chrono::duration<int, std::ratio<kSecondsInWeek>>{1}};
+  vector<std::chrono::minutes> overdue_per_severity_;
```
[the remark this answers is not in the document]
HNOONa-0: Memory is so small, its the same size of `NUM_SEVERITIES`, which is 4 and the size of the vector is static.

I thought this would be more efficient and simpler but I can make those changes.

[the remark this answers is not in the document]
sergiud: You are thinking about desktop systems where you have plenty of memory to spare. glog, however, is also used on embedded hardware where you cannot afford allocating memory and leave it unused.

Instead of an `std::vector` you could also use `std::array`. This would avoid heap allocation.

HNOONa-0:
Hey @sergiud, I initially made the wrong assumptions, I thought log files must have severity in their names, but that’s not the case. I have added methods to enable, disable logging for both types of files (with and without severity in their names) and exported those functions. I also added a test file that checks for deferred cleanup logic, but I haven’t integrated it with CMake yet.

PR #1146, opened 2024-12-19
Add LWP id to dump info

stdpain:
LWP can help gdb find the corresponding thread more quickly.

```
*** SIGSEGV (@0xff) received by PID 2699683 (TID 0x7f342020a640 LWP 2700477) from PID 255; stack trace: ***
    @     0x7f3501070ee8 (/usr/lib/x86_64-linux-gnu/libc.so.6+0x99ee7)
```

### DIFF
--- a/src/signalhandler.cc
+++ b/src/signalhandler.cc
@@ -55,6 +55,10 @@
 #ifdef HAVE_UNISTD_H
 #  include <unistd.h>
 #endif
+#if defined(HAVE_SYS_SYSCALL_H) && defined(HAVE_SYS_TYPES_H)
+#  include <sys/syscall.h>
+#  include <sys/types.h>
+#endif
 
 namespace google {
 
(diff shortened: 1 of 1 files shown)


src/signalhandler.cc
```diff
@@ -218,8 +220,14 @@ void DumpSignalInfo(int signal_number, siginfo_t* siginfo) {
   formatter.AppendString(oss.str().c_str());
 
   formatter.AppendString(") ");
+
   // Only linux has the PID of the signal sender in si_pid.
 #  ifdef GLOG_OS_LINUX
+  pid_t tid = syscall(SYS_gettid);
+  formatter.AppendString("(LWP ");
```
sergiud, line 227: Please surround the LWP by previously opened braces.

src/signalhandler.cc
```diff
@@ -53,6 +53,8 @@
 #  include <sys/ucontext.h>
 #endif
 #ifdef HAVE_UNISTD_H
+#  include <sys/syscall.h>
```
sergiud, line 56: The availability of a POSIX header (`unistd.h`) does not imply the availability of LInux headers. Please use the appropriate config macros (i.e., `HAVE_SYS_SYSCALL_H`, etc.) here.

sergiud (changes requested):

Please see my remarks.

src/signalhandler.cc
```diff
@@ -216,8 +220,13 @@ void DumpSignalInfo(int signal_number, siginfo_t* siginfo) {
   std::ostringstream oss;
   oss << std::showbase << std::hex << std::this_thread::get_id();
   formatter.AppendString(oss.str().c_str());
-
+#if defined (HAVE_SYS_SYSCALL_H) && defined (HAVE_SYS_TYPES_H)
+  pid_t tid = syscall(SYS_gettid);
```
sergiud, line 224: Unfortunately, using `syscall` triggers the following error on macOS caused by a deprecation warning:
```console
src/signalhandler.cc:224:15: error: 'syscall' is deprecated: first deprecated in macOS 10.12 - syscall(2) is unsupported; please switch to a supported interface. For SYS_kdebug_trace use kdebug_signpost(). [-Werror,-Wdeprecated-declarations]
  pid_t tid = syscall(SYS_gettid);
```
We try to avoid deprecated APIs.

stdpain: change to: print LWP id only on Linux platforms

sergiud:
Thanks!

sergiud (approved):

PR #1151, opened 2025-02-03
Optimize wildcard matching by converting SafeFNMatch_ to an iterative approach

RadoslavPetkow:
Refactored SafeFNMatch_
Replaced the recursive implementation with an iterative approach to handle ‘*’ and ‘?’ wildcards. This change eliminates recursive call overhead and reduces potential stack usage, leading to improved performance in logging scenarios.
Maintained Behavior
The change preserves all existing semantics, ensuring that only the internal implementation of wildcard matching is affected.
Minor Code Cleanups
Made slight adjustments to pointer arithmetic and code formatting in functions such as VLOG2Initializer for improved clarity and robustness.

### DIFF
--- a/src/vlog_is_on.cc
+++ b/src/vlog_is_on.cc
@@ -36,6 +36,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <limits>
 #include <mutex>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


sergiud (changes requested):
Thanks for the PR.

It looks like the changes cause the demangle unit test to no longer pass on Linux. Could you please also split the `SafeFNMatch_` unrelated refactorings and submit these in a separate PR?

RadoslavPetkow:
> Thanks for the PR.
> 
> It looks like the changes cause the demangle unit test to no longer pass on Linux. Could you please also split the `SafeFNMatch_` unrelated refactorings and submit these in a separate PR?

Hey, no worries. I’m really sorry for the delay—I’ve been swamped with work at university lately. I’ll take a closer look at the demangle unit test issue on Linux, and I’ll also split the SafeFNMatch_ refactorings into a separate PR as requested. Thanks for your patience and the feedback!

PR #1159, opened 2025-03-04
chore(security): avoid use of `strcpy`

Saadnajmi:
Internally, we use this library with React Native. We noticed that Xcode's static analysis would point out a potentially unsafe call to `strcpy`. We are doing a proper bound check, but to be safer, we could also use something like `strncpy` / `strncpy_s` / `strlcpy` / `memcpy` to handle this case. Of those, the latter felt the most portable, so I went with that.

### DIFF
--- a/src/symbolize.cc
+++ b/src/symbolize.cc
@@ -832,8 +832,10 @@ static ATTRIBUTE_NOINLINE bool SymbolizeAndDemangle(
   Dl_info info;
   if (dladdr(pc, &info)) {
     if (info.dli_sname) {
-      if (strlen(info.dli_sname) < out_size) {
-        strcpy(out, info.dli_sname);
+      int name_length = strlen(info.dli_sname);
+      if (name_length < out_size) {
+        memcpy(out, info.dli_sname, name_length);
+        out[name_length] = '\0';
         // Symbolization succeeded.  Now we try to demangle the symbol.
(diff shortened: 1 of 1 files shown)


Saadnajmi:
Signed the CLA

sergiud (changes requested):
Thanks for the PR.

Silencing the static analyzer by reimplementing `strcpy` is not meaningful. Given the bounds are checked I don't understand what the current changes actually fix.

Saadnajmi:
> Thanks for the PR.
> 
> 
> 
> Silencing the static analyzer by reimplementing `strcpy` is not meaningful. Given the bounds are checked I don't understand what the current changes actually fix.

Fair. I would have preferred to use something like 'strlcpy' which Xcode actually suggested, but I wasn't sure that would be portable enough, as I'm not sure what Google's C / C++ standards are. Would you have an idea?

Saadnajmi:
@sergiud Would `strlcpy` work? 
w.r.t what this actually fixes, it's mostly me not wanting to maintain a fork of glog / glog with a patch for this one line. I understand Xcode / clang is over-reporting here since the usage is actually sage, but 🤷‍♂️

sergiud:
@Saadnajmi I suggest we test this in [ng-log](https://github.com/ng-log/ng-log) that supersedes glog which is being deprecated.

Please open a PR there.

PR #1164, opened 2025-05-14
add check and set RTTI macros

siyuhong:
I noticed that ``DISABLE_RTTI`` was used in ``src/logging.cc``, but ``DISABLE_RTTI`` was not set. 
If RTTI is disabled at compile time (for example,`` g++ -fno-rtti``), it will cause an error.


ukai:
as README says, glog is deprecated, and no longer maintained.

siyuhong:
> as README says, glog is deprecated, and no longer maintained.

Ok, I'm very sorry I didn't check ``README``.

